PR #1, opened 2015-07-05
voice verification and bugfix

jooray:
* added option to perform voice verification instead of SMS
 * fixed a bug when signalingKey would not be generated until verification, but it's needed for load() to perform without error (fresh registration did not work without this).


AsamK:
Thanks very much! I cherry-picked the voice verification commit.
I fixed the signalingKey issue differently now.

PR #4, opened 2015-10-06
Make Json store use Jackson instead of Gson (as it's already linked)

xardazz:
Added Jackson serializtion/deserialization support.


AsamK:
Thanks, looks good!

PR #8, opened 2016-03-27
URL should point to textsecure-service subdomain

nrizzio:
10719a443a88d06ef5734f0e17f71316b1473edf introduced a new bug where the CLI won't connect to the Signal service, since https://SignalService-service.whipsersystems.org isn't the correct hostname. This fixes the problem by changing the URL back to what it should be.


AsamK:
Thanks, I had already fixed that locally, but not pushed yet.

PR #10, opened 2016-04-11
Receive messages

individual-it:
I was looking in the code for a way to receive messages and added the command to the documentation


AsamK:
Thanks.
It's already contained in this line "usage: signal-cli [-h] [-u USERNAME] [-v] {register,verify,send,quitGroup,updateGroup,receive} ..." but probably not very obvious to find.

PR #11, opened 2016-05-18
Added quit group option

lucaconte:
as updateGroup sub option I've added the quit option that give a user the opportunity to leave a subscribed group


lucaconte:
mistake: found quitGroup

PR #31, opened 2016-11-16
Update README.md

larsborn:
I use a server monitoring solution (Zabbix) and wanted to use signal-cli for notifications. The monitoring solutions can call external scripts in case of problems but obviously uses it's own user to do this. I read about --config in the source and thought, that other people would benefit from this information as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,12 @@ usage: signal-cli [-h] [-v] [--config CONFIG] [-u USERNAME | --dbus | --dbus-sys
 
           signal-cli -u USERNAME trust -a NUMBER
 
+* Set configuration directory
+
+          signal-cli -config=/home/other_user/.config/signal
+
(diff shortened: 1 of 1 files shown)


AsamK:
Thanks, there's a typo in the PR, it's --config, not -config. The --config is already included in the first line in the usage section, although not explained.
Also the other user needs write access to the config files, to update the session store and known keys.
The recommended way for this scenario would be to have a, possibly dbus activated, signal-cli service with its own user. Then the sender doesn't need access to the config files, but can use the dbus api.

larsborn:
all right, I fixed the type.

I didn't dare to use the dbus interface because it was marked as "experimental". Or are there plans to change this in the near future?

AsamK:
The dbus interface is currently very minimalistic, and I'm a bit reluctant to declare it stable. However I have no immediate plans for changes to the interface.

larsborn:
All right, good to know. I'll probably give it a try then soon!

PR #34, opened 2016-11-27
use java7's nio api to get more detailed error messages

Trolldemorted:
file.delete will not yield any useful information - Files.delete does, but requieres java7+.


AsamK:
thanks, please add spaces around the + sign.
signal-cli requires Java 7 already, so that's not a problem.

Trolldemorted:
sorry, i have amended my commit

PR #52, opened 2017-02-12
Add the packaging metadata to build the signal-cli snap

elopio:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds that then can be distributed to many users through the Ubuntu store.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,23 @@
+name: signal-cli
+version: master
+summary: commandline interface for libsignal-service-java
+description: |
+  It supports registering, verifying, sending and receiving messages.
+  It is primarily intended to be used on servers to notify admins of important
+  events.
+
+grade: devel
+confinement: strict
+
+apps:
+  signal-cli:
+    command: signal-cli/bin/signal-cli --config $SNAP_USER_DATA
+    plugs: [network, network-bind, home]
+
+parts:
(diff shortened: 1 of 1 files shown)


elopio:
Hello!

I've just found your project and I think it's awesome. First, I'm trying to move all my communications to a remote server. A cli for signal is something I was missing, so thank you. And the second reason is that we are developing snapcraft, and it has a basic gradle plugin. We need more users of this plugin to make it better, so any feedback you can give us will be greatly appreciated :)

You can find more information about snapcraft here: http://snapcraft.io/

To test this in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/signal-cli
    $ cd signala-cli
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous

(dangerous because the snap you build yourself doesn't come signed from the trusted store. Once you push it to the store, your users will just have to do $ sudo snap install signal-cli)

If you have any questions or comments, please let me know.

elopio:
Ping @AsamK. Any chance you can give this a review?

AsamK:
I'm not planning to publish snaps, but if someone else wants to do it, feel free to add a link to the wiki page: https://github.com/AsamK/signal-cli/wiki/Binary-distributions

PR #60, opened 2017-02-19
Add eclipse plugin to build.gradle.

pimotte:
This allows easy generation of eclipse project files using `./gradlew
eclipse`

Exactly what it says on the tin. I like working with eclipse, and then this is really useful.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ build/
 *.swp
 *.iml
 local.properties
+.classpath
+.project
+.settings/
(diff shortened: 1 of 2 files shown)


pimotte:
To be a little more specific why I'd like to get this in: Doing this allows gradle to pull the right dependencies into the right place for eclipse. I don't think this should screw anything up for non-eclipse users, but if you have any qualms I'm happy to hear them.

PR #61, opened 2017-02-19
Get contact and group names over dbus

thefinn93:
This pull request creates two dbus new methods, `getGroupName` and `getContactName`, which return the names of groups and contacts.


thefinn93:
Sorry for the spam, I opened this not expecting to do more, then i did more, so #62 came into being.

PR #62, opened 2017-02-20
Allow retreving and updating group info and contact names via dbus

thefinn93:
This PR enables contact names, group names and group member lists to be read and updated over dbus. Superseeding #61 which just allows reading.

### DIFF
--- a/src/main/java/org/asamk/signal/Manager.java
+++ b/src/main/java/org/asamk/signal/Manager.java
@@ -834,8 +834,7 @@ class Manager implements Signal {
         if (!avatar.isEmpty()) {
             optAvatar = avatar;
         }
-        System.err.println("sendUpdateGroupMessage(" + Base64.encodeBytes(groupId) + ", " + optName + ", " + optMembers + ", " + optAvatar + ");");
-        sendUpdateGroupMessage(groupId, optName, optMembers, null);
+        sendUpdateGroupMessage(groupId, optName, optMembers, optAvatar);
     }
 
     private void requestSyncGroups() throws IOException {
--- a/src/main/java/org/asamk/signal/Manager.java
+++ b/src/main/java/org/asamk/signal/Manager.java
@@ -822,19 +822,16 @@ class Manager implements Signal {
 
     @Override
     public void updateGroup(byte[] groupId, String name, List<String> members, String avatar) throws IOException, EncapsulatedExceptions, GroupNotFoundException, AttachmentInvalidException {
-        String optName = null;
-        Collection<String> optMembers = null;
-        String optAvatar = null;
-        if (!name.isEmpty()) {
-            optName = name;
+        if (name.isEmpty()) {
+            name = null;
         }
-        if (members.size() > 0) {
-            optMembers = members;
+        if (members.size() == 0) {
+            members = null;
         }
-        if (!avatar.isEmpty()) {
-            optAvatar = avatar;
+        if (avatar.isEmpty()) {
+            avatar = null;
         }
-        sendUpdateGroupMessage(groupId, optName, optMembers, optAvatar);
+        sendUpdateGroupMessage(groupId, name, members, avatar);
     }
 
     private void requestSyncGroups() throws IOException {
--- a/src/main/java/org/asamk/signal/Manager.java
+++ b/src/main/java/org/asamk/signal/Manager.java
@@ -802,12 +802,22 @@ class Manager implements Signal {
 
     @Override
     public String getGroupName(byte[] groupId) {
-        return getGroup(groupId).name;
+        GroupInfo group = getGroup(groupId);
+        if (group == null) {
+            return "";
+        } else {
+            return group.name;
+        }
     }
 
     @Override
     public List<String> getGroupMembers(byte[] groupId) {
-        return new ArrayList<String>(getGroup(groupId).members);
+        GroupInfo group = getGroup(groupId);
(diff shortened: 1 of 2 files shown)


src/main/java/org/asamk/signal/Manager.java
```diff
@@ -775,6 +775,59 @@ public void sendEndSessionMessage(List<String> recipients) throws IOException, E
         sendMessage(messageBuilder, recipients);
     }
 
+    @Override
+    public String getContactName(String number) {
+        ContactInfo contact = contactStore.getContact(number);
+        if(contact == null) {
+            return number;
```
AsamK: I think it's better to return an empty string here, to indicate, that the contact is unknown

src/main/java/org/asamk/signal/Manager.java
```diff
@@ -775,6 +775,59 @@ public void sendEndSessionMessage(List<String> recipients) throws IOException, E
         sendMessage(messageBuilder, recipients);
     }
 
+    @Override
+    public String getContactName(String number) {
+        ContactInfo contact = contactStore.getContact(number);
+        if(contact == null) {
```
AsamK: Please add a space after the if -> "if ("
also for the other ifs added in this PR.

src/main/java/org/asamk/signal/Manager.java
```diff
@@ -775,6 +775,59 @@ public void sendEndSessionMessage(List<String> recipients) throws IOException, E
...
+            System.out.println("Updating contact " + number + " name " + contact.name + " -> " + name);
+        }
+        contact.name = name;
+        contactStore.updateContact(contact);
+        save();
+    }
+
+    @Override
+    public String getGroupName(byte[] groupId) {
+        return getGroup(groupId).name;
```
AsamK: The group can be null, please check this (like in getContactName)

src/main/java/org/asamk/signal/Manager.java
```diff
@@ -775,6 +775,59 @@ public void sendEndSessionMessage(List<String> recipients) throws IOException, E
...
+    }
+
+    @Override
+    public List<String> getGroupMembers(byte[] groupId) {
+        return new ArrayList<String>(getGroup(groupId).members);
```
AsamK: The group can be null, please check this

src/main/java/org/asamk/signal/Manager.java
```diff
@@ -775,6 +775,59 @@ public void sendEndSessionMessage(List<String> recipients) throws IOException, E
...
+            optName = name;
+        }
+        if(members.size() > 0) {
+            optMembers = members;
+        }
+        if(!avatar.isEmpty()) {
+            optAvatar = avatar;
+        }
+        System.err.println("sendUpdateGroupMessage(" + Base64.encodeBytes(groupId) + ", " + optName + ", " + optMembers + ", " + optAvatar + ");");
+        sendUpdateGroupMessage(groupId, optName, optMembers, null);
```
AsamK: null -> optAvatar

src/main/java/org/asamk/signal/Manager.java
```diff
@@ -775,6 +775,59 @@ public void sendEndSessionMessage(List<String> recipients) throws IOException, E
...
+        if(!name.isEmpty()) {
+            optName = name;
+        }
+        if(members.size() > 0) {
+            optMembers = members;
+        }
+        if(!avatar.isEmpty()) {
+            optAvatar = avatar;
+        }
+        System.err.println("sendUpdateGroupMessage(" + Base64.encodeBytes(groupId) + ", " + optName + ", " + optMembers + ", " + optAvatar + ");");
```
AsamK: For consistency: "Updating group GROUP_ID, name NAME, ...

src/main/java/org/asamk/signal/Manager.java
```diff
@@ -775,6 +775,59 @@ public void sendEndSessionMessage(List<String> recipients) throws IOException, E
...
+    @Override
+    public List<String> getGroupMembers(byte[] groupId) {
+        return new ArrayList<String>(getGroup(groupId).members);
+    }
+
+    @Override
+    public void updateGroup(byte[] groupId, String name, List<String> members, String avatar) throws IOException, EncapsulatedExceptions, GroupNotFoundException, AttachmentInvalidException {
+        String optName = null;
+        Collection<String> optMembers = null;
+        String optAvatar = null;
```
AsamK: Is there a special reason to use separate variables?
You can just set the method arguments to null, if they're empty:
```
if (name != null && name.isEmpty()) {
    name = null;
}
```

thefinn93: I tried that initially and it didn't seem to like it. I think mostly my inexperience with Java. This made it work... I'll dig into it, this way is pretty ugly.

src/main/java/org/asamk/signal/Manager.java
thefinn93: oops, didn't realize I left this in there at all, I was just using it for debugging. Would you rather I simply remove it?

src/main/java/org/asamk/signal/Manager.java
thefinn93: Ahh, I remember now why that happened. Was due to some other issue caused by my system config.

thefinn93:
All requested changes have been made (with the exception of the system.err.println which I simply removed)

Issue #58, opened 2017-02-18
remove documentation from README


anarcat:
as discussed in #57, the README is now out of date. we should probably move all of that documentation to the manpage instead.

furthermore, it seems to me it would be better if the manpage was rendered directly on github - for this i believe it would be sufficient to rename it to `.asciidoc`, see the [asciidoc readme for an example](https://github.com/asciidoc/asciidoc/blob/master/README.asciidoc).

pimotte:
Somewhat related: I can't seem to find how to get a group id in the first place. This might be a nice addition to the resulting asciidoc

Trolldemorted:
@pimotte open signal-cli's config and ctrl-f for `groupId`

AsamK:
The group id is shown when creating a group and when receiving a group message.
An additional listGroups command might be a good idea, to list all known groups with id and members.

AsamK:
Done in 0744dcccf117064dcf6c8dbb1312565c0f3e15cb
The README now contains only basic information and links to the man page and the wiki.

Some thoughts, on how the documentation should be handled now:
* README: basic information, how to build, how to run first time, ...
* man page: Description of every command/parameter, some examples
* wiki: More detailed explanations and howtos

PR #63, opened 2017-02-20
Add ListGroups command

pimotte:
Option: -d/--detailed to display group members

As discussed in #58, this might be useful. 

Happy to hear any and all feedback:)

### DIFF
--- a/src/main/java/org/asamk/signal/Main.java
+++ b/src/main/java/org/asamk/signal/Main.java
@@ -503,9 +503,10 @@ public class Main {
                     }
     
                     List<GroupInfo> groups = m.getGroups();
+                    boolean detailed = ns.getBoolean("detailed");
     
                     for (GroupInfo group : groups) {
-                        printGroup(group, ns.getBoolean("detailed"));
+                        printGroup(group, detailed);
                     }
                     break;
                 case "listIdentities":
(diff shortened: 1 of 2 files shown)


src/main/java/org/asamk/signal/Main.java
```diff
@@ -492,6 +492,22 @@ public void handle(Signal.MessageReceived s) {
...
+                    }
+                    if (!m.isRegistered()) {
+                        System.err.println("User is not registered.");
+                        return 1;
+                    }
+    
+                    List<GroupInfo> groups = m.getGroups();
+    
+                    for (GroupInfo group : groups) {
+                        printGroup(group, ns.getBoolean("detailed"));
```
AsamK: Please create a variable `boolean detailed` outside the for loop for this.

src/main/java/org/asamk/signal/Manager.java
```diff
@@ -580,6 +580,10 @@ private GroupInfo getGroupForSending(byte[] groupId) throws GroupNotFoundExcepti
         }
         throw new NotAGroupMemberException(groupId, g.name);
     }
+    
+    public List<GroupInfo> getGroups() {
+    	return groupStore.getGroups();
```
AsamK: Fix indentation, tab -> spaces

src/main/java/org/asamk/signal/Main.java
```diff
@@ -622,6 +638,17 @@ private static void printIdentityFingerprint(Manager m, String theirUsername, Js
         System.out.println(String.format("%s: %s Added: %s Fingerprint: %s Safety Number: %s", theirUsername,
                 theirId.trustLevel, theirId.added, Hex.toStringCondensed(theirId.getFingerprint()), digits));
     }
+    
+    private static void printGroup(GroupInfo group, boolean detailed) {
+        System.out.println(String.format("Group id: %s\n Group name: %s \n active: %s",
```
AsamK: I'd prefer having each group in a single line, e.g.:
"Id: %s Name: \"%s\" Active: %s Members: ..."

pimotte:
Fixed all of those:)

PR #67, opened 2017-02-23
Refactoring into packages

pimotte:
This is a small re-organization of the code. The reason I did this is that I'd like to extract some of the logic currently in Main.java and perhaps the manager, and to keep the overview I wanted to divide the current classes into coherent packages. 

The commits are split in 1) the move itself and 2) making some methods public to allow cross-package usage.

Happy to hear any and all feedback:)


src/main/java/org/asamk/signal/Manager.java
```diff
@@ -16,18 +16,70 @@
  */
 package org.asamk.signal;
 
-import com.fasterxml.jackson.annotation.JsonAutoDetect;
```
AsamK: Please avoid moving around the import statements. I suppose this was an automatic change from your IDE.
It makes the diff difficult to read and might cause unnecessary merge conflicts later.

pimotte:
Renamed thread->threads and fixed the import order. (I still included the expansions of *, I'm aware this yields a crazy long list, but I'll probably try to split Manager into smaller classes at some point.)

Issue #40, opened 2016-12-17
How to safely capture output?


kb100:
The format that signal-cli gives output makes it very hard for a script to safely capture message content.

E.g. by sending a message like:
"Hey! Nothing suspicious about this message.
Attachments: 
- image/jpeg (Pointer)
  Id: XXXXXXXXXXXXXXX Key length: 64 Relay: 
  Size: nnnnnn bytes
  Stored plaintext in: /path/to/secret/key"
"

A malicious user can trick a script that is grepping for the given format to perhaps act on a file guessed by the malicious user.
I suggest optional formatting command line parameters:

1 --json, output is guaranteed to be valid json

2 -0 -Z or --null (see man grep and man xargs). use zero bytes as delimeters, e.g. to delineate when message text ends.

AsamK:
That's a known flaw of the current output. Sofar I've only used the output for testing, not scripting. Currently the only safe way to read messages is via the (limited) dbus api.
A json output format is a good idea.

mpex:
+1 for json output

PR #71, opened 2017-02-28
Add a JSON receive logger for #40

jackschmidt:
Here is one way to add JSON logging, perhaps in answer to #40 . I think it is fairly thorough in decoding the messages. It doesn't do special decoding for receiving the list of groups or contacts attachments. If you like the idea of decoding those, I made #72 too.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,7 @@ dependencies {
     compile 'org.bouncycastle:bcprov-jdk15on:1.55'
     compile 'net.sourceforge.argparse4j:argparse4j:0.7.0'
     compile 'org.freedesktop.dbus:dbus-java:2.7.0'
+    compile 'org.json:json:20160810'
 }
 
 jar {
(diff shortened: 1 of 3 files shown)


kb100:
It seems to me that whether signal-cli should output json formatted data or not should be a suboption of receive, just like receive attachments. I would expect to be able to use it like 

      signal-cli -u USERNAME  receive --ignore-attachments --json

Additionally, this PR seems to introduce a whole new logging framework. While a logging framework may be added in the future, I don't think it should be combined into this PR. My suggestion for getting this PR merged would be the following:

1. Make --json a suboption of receive, the receiveMessages functions need to have this new option added and handled there.
2.  Remove the logging framework. Instead output the json to stdout (find where we were going to System.out.print(plaintext) and instead print the json if the json flag is set.)

AsamK:
Thanks for your effort, I've finished my own json output format now. That uses the jackson json library, that's already used in libsignal and is available via `receive --json`: 8717665d1d273a32afef136c15c0d5abaaae0f85

PR #90, opened 2017-06-20
jsonevtloop mode + perl module

kbin76:
Hi,
what do you think of this? 

I've removed all my json generation (also removed dependencies on gson library) in favour of what you pushed the other day. 

I also added some convenience functionality to your json data output, for instance now it always emits additional timestamp in readable ISO utc format and the groups name if it's a group message. The latter which in turn resulted in the need for having Manager object in creation of JsonGroupInfo object.

However I felt that this wasn't properly belonging under "daemon" option, so it's still invoked using it's own special "jsonevtloop" command.

Currently the JSON output is quite more complex than what is expected as JSON input, I'v prefer if the input would mirror the output in terms of JSON schema, but otoh that would be uneccessary complex (keep it simple).

Perl module is under perl/ with a couple of example scripted bots.

BR
Kalle

### DIFF
--- a/src/main/java/org/asamk/signal/Main.java
+++ b/src/main/java/org/asamk/signal/Main.java
@@ -49,9 +49,7 @@ import static org.whispersystems.signalservice.internal.util.Util.isEmpty;
 public class Main {
 
     public static void main(String[] args) {
-        // Register our own security provider
-        Security.insertProviderAt(new SecurityProvider(), 1);
-        Security.addProvider(new BouncyCastleProvider());
+        installSecurityProviderWorkaround();
 
         Namespace ns = parseArgs(args);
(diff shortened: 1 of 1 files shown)


thefinn93:
Is there something preventing this from getting merged? I'd love to build on top of it, but not if it's going to undergo changes before merge.

skorokithakis:
I realize this is pretty old, but would you mind merging the current master into your branch, @kbin76, so we can use your fork? I really need this functionality but this PR hasn't gotten any replies.

/cc @AsamK

kbin76:
I closed this pull request and created a new one where I have re-implemeted this functionality. It was not possible to merging the current master into my old branch, so i had to start from scratch again (more or less).

PR #94, opened 2017-07-04
Added Dockerfile and instructions for the Docker image.

ksylvan:
Integrate with Docker hub automated build to provide a small alpine linux based docker image.

This makes it possible to run `signal-cli` immediately on any machine where Docker is available.

See https://hub.docker.com/r/kayvan/signal-cli/ for the current image based on my fork.

I've set up the automated build to generate tagged images based on the git release tags.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+# Dockerfile for automated build of signal-cli
+#
+# Refer to the signal-cli github pages for detailed Usage.
+#
+FROM gradle:3.5-jdk7-alpine
+
+LABEL maintainer "Kayvan Sylvan <kayvansylvan@gmail.com>"
+
+COPY . /tmp/src
+WORKDIR /tmp/src
+
+USER root
+
+RUN ./gradlew build && ./gradlew installDist && ./gradlew distTar \
+  && tar xf build/distributions/signal-cli-*.tar -C /opt \
+  && ln -sf /opt/signal-cli-*/bin/signal-cli /usr/local/bin/ \
+  && rm -rf /tmp/src
+
+USER gradle
+WORKDIR /home/gradle
+
+ENTRYPOINT ["/usr/local/bin/signal-cli"]
--- a/README.md
+++ b/README.md
@@ -7,7 +7,32 @@ signal-cli is primarily intended to be used on servers to notify admins of impor
 
 ## Installation
 
-You can [build signal-cli](#building) yourself, or use the [provided binary files](https://github.com/AsamK/signal-cli/releases/latest), which should work on Linux, macOS and Windows. For Arch Linux there is also a [package in AUR](https://aur.archlinux.org/packages/signal-cli/). You need to have at least JRE 7 installed, to run signal-cli.
+You can [build signal-cli](#building) yourself, run the
+[auto-built Docker image](https://hub.docker.com/r/kayvan/signal-cli/),
+or use the [provided binary files](https://github.com/AsamK/signal-cli/releases/latest), which should work on Linux, macOS and Windows. For Arch Linux there is also a [package in AUR](https://aur.archlinux.org/packages/signal-cli/). You need to have at least JRE 7 installed, to run signal-cli.
+
+## Running the Docker image
+
+Where you have a working Docker installation, you can run signal-cli like this:
+
+    docker run -v $HOME:/home/gradle --rm -it kayvan/signal-cli -h
+
+Use the `-v $HOME:/home/gradle` arguments to ensure that your signal-cli
+settings and data end up in `$HOME/.config/signal/` and persist.
+
+You can link the CLI client to your primary device (phone, tablet, etc.):
+
+    docker run -v $HOME:/home/gradle --rm -it kayvan/signal-cli link
+    tsdevice:/?uuid=SOME_ID&pub_key=SECRET
+
+Use qrencode to turn the generated `tsdevice:` URL into a QR code to scan:
+
+    qrencode -t ANSI256 'tsdevice:/?uuid=SOME_ID&pub_key=SECRET'
+
+When you scan the code using your master device, signal-cli will emit a
+verification message:
+
+    Associated with: [YourPhoneNumber]
 
 ### Install system-wide on Linux
 See [latest version](https://github.com/AsamK/signal-cli/releases).


Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Dockerfile for automated build of signal-cli
+#
+# Refer to the signal-cli github pages for detailed Usage.
+#
+FROM gradle:3.5-jdk7-alpine
+
+LABEL maintainer "Kayvan Sylvan <kayvansylvan@gmail.com>"
+
+USER root
```
anarcat: why run the whole build as root? shouldn't just the installation be performed with elevated privileges?

ksylvan: That's exactly what happens. Only the build happens as root, to install the needed components in the image. Further down in the Dockerfile we change the user to `gradle` before running `signal-cli`

When you run a container, it runs as the `gradle` user in the container but you can always change that on the command line.

anarcat: i think you misunderstood the point i was raising. i was suggesting to run the build as a non-root user and run the *install* as root only.

i do see that signal itself run as gradle, and that's fine. :) i just wonder if the build part of the install process should run as gradle as well.

ksylvan: I hear you. I couldn't figure out an easy way to do that, and given that the install happens inside a `gradle:jdk7-alpine` container and does not touch the host on which you are running, I don't think it matters.

The alpine container is a minimal `busybox` based system. What to you propose as the way to run the build as a non-root user?

ksylvan:
I slightly reworked the branch, leaving the running java process's `WORKDIR` as `/home/gradle` instead of the (at that point non-existent) source directory.

ksylvan:
In a further answer to "why run the whole build as root? shouldn't just the installation be performed with elevated privileges?" - That is where I started, and this is the error I got, which I worked around by running the build as root.

```
$ docker build -t test .                           
Sending build context to Docker daemon  1.455MB
Step 1/10 : FROM gradle:3.5-jdk7-alpine
 ---> 2b28527e76c5
Step 2/10 : LABEL maintainer "Kayvan Sylvan <kayvansylvan@gmail.com>"
 ---> Using cache
 ---> 520deedee141
Step 3/10 : COPY . /tmp/src
 ---> f54e0a798e0f
Removing intermediate container af82e9c8cde3
Step 4/10 : WORKDIR /tmp/src
 ---> 5eb8637e9a46
Removing intermediate container 5bbc6beb9025
Step 5/10 : RUN ./gradlew build && ./gradlew installDist && ./gradlew distTar
 ---> Running in 4ac3ad3cea31
Downloading https://services.gradle.org/distributions/gradle-3.5-bin.zip
Unzipping /home/gradle/.gradle/wrapper/dists/gradle-3.5-bin/daoimhu7k5rlo48ntmxw2ok3e/gradle-3.5-bin.zip to /home/gradle/.gradle/wrapper/dists/gradle-3.5-bin/daoimhu7k5rlo48ntmxw2ok3e
Set executable permissions for: /home/gradle/.gradle/wrapper/dists/gradle-3.5-bin/daoimhu7k5rlo48ntmxw2ok3e/gradle-3.5/bin/gradle
Starting a Gradle Daemon (subsequent builds will be faster)

FAILURE: Build failed with an exception.

* Where:
Build file '/tmp/src/build.gradle' line: 1

* What went wrong:
A problem occurred evaluating root project 'signal-cli'.
> Failed to apply plugin [class 'org.gradle.api.plugins.BasePlugin']
   > Could not create service of type CachingFileHasher using TaskExecutionServices.createFileSnapshotter().

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 49.082 secs

FAILURE: Build failed with an exception.

* What went wrong:
Could not update /tmp/src/.gradle/3.5/file-changes/last-build.bin

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
The command '/bin/sh -c ./gradlew build && ./gradlew installDist && ./gradlew distTar' returned a non-zero code: 1
```

ksylvan:
@anarcat any other concerns/questions?

ksylvan:
Can this get merged?

It simply enables setting up an automated Docker build (like the one sitting here (based on my fork) https://hub.docker.com/r/kayvan/signal-cli/ that automatically tracks versions of `signal-cli` in a small and ready-to-run package.

ksylvan:
Any updates on this?

ksylvan:
Ping.

anarcat:
pong... don't take my comments as blockers. :)

ksylvan:
So who does the merge?

AuspeXeu:
Any update on this PR ? :)

ksylvan:
Ping

pgaxatte:
Any news on this? Can it be merged?

Dockerfile
francisg-gc: there's no point in running / not worth it as non-root as it's "literally" (practically) the same as a non-root user if the container is not privileged as the attack surface is mostly the same, the efforts are better spent elsewhere ;)

francisg-gc:
@anarcat says this is not blocked

could any of you please merge ? :D 

@asamk @pimotte @riamse @kllp @chr1shaefn3r @trolldemorted @thisisparker

thank you!

Dockerfile
anarcat: > there's no point in running / not worth it as non-root as it's "literally" (practically) the same as a non-root user if the container is not privileged as the attack surface is mostly the same, the efforts are better spent elsewhere ;)

it's definitely not the same. definitely not "literally" the same, at least, and not even *practically* the same either. but whatever, as i said it's not a blocker.

francisg-gc: hehehe :D i think we're in agreement here despite my confusing previous message :D

i guess i should've pointed out that at the end of the day, to break out of the container it's gonna be the kernel the biggest attack surface upon which uid=0 wont make much difference to exploit if the container is not privileged. arguably uid=0 is more dangerous when the container is privileged than otherwise :)

also this Dockerfile could have separate build and run environments in the same file, a refresh of it perhaps is in order

LoveIsGrief:
For anybody looking for an image, I set up https://hub.docker.com/r/bisentenialwrug/signal-cli . It's automatically updated with changes from upstream

alexanderadam:
> It's automatically updated with changes from upstream

Nice. Did you use GitHub actions for it? Do you have any documentation about the setup?

LoveIsGrief:
@alexanderadam I moved to gitlab. Gitlab support hourly pushes and pulls from a remote repo (aka mirroring). Before, I used to merge `master` into my branch (`dockerfile`) manually. That is then be pushed by gitlab to github, which triggers a build on dockerhub.

Now I have a [dirty script `f2g`](https://pypi.org/project/fork2gitlab/).
That script is now in an hourly cron and does the job of merging `master` into `dockerfile`. If there are no merge conflics it simply pushes to the gitlab repo. Merge conflicts create a merge request on gitlab.

alexanderadam:
Awesome, thank you Michael (for the image and the explanation)! :pray:

jace:
@LoveIsGrief https://hub.docker.com/r/bisentenialwrug/signal-cli is showing as last updated two years ago, while signal-cli's last release was this month. Looks like the automated releases are no longer working.

LoveIsGrief:
@jace you're right, thanks. I wasn't using it anymore. Looks like there was a force-push to master that broke the sync... I'm updating the `Dockerfile` and fixing the git inconsistencies wrought upon by the force-push.

AsamK:
I'm not planning to add a Dockerfile to the repo, but I've created a wiki page, which links to morph027's repo: https://github.com/AsamK/signal-cli/wiki/Binary-distributions

PR #102, opened 2017-09-21
Add getGroupIds() to DBUS interface

cholin:
`getGroupIds()` returns a list of group ids (byte array)

This comes handy if you want to use `getGroup*()` methods but don't have the group id.


axapta:
Are there plans to include this functionality in the near future @AsamK, because I would like to use this functionality in my bot?

AsamK:
thanks, sorry for the delay

Issue #127, opened 2018-05-05
JSON support for daemon


Riamse:
It'd be cool if `signal-cli -u ... daemon` could print in json, like how `signal-cli -u ... receive --json` does it right now.

PR #128, opened 2018-05-08
Dbus json

Riamse:
Duct tape solution to issue #127 . The real solution would be to rearchitect the message handlers, but hey, it's easy and it works.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,6 +38,8 @@ jar {
 
 run {
     if (project.hasProperty("appArgs")) {
+        // allow passing command-line arguments to the main application e.g.:
+        // $ gradle run -PappArgs="['-u', '+...', 'daemon', '--json']"
         args Eval.me(appArgs)
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -36,6 +36,12 @@ jar {
     }
 }
 
+run {
+    if (project.hasProperty("appArgs")) {
+        args Eval.me(appArgs)
+    }
+}
+
 // Find any 3rd party libraries which have released new versions
 // to the central Maven repo since we last upgraded.
 // http://daniel.gredler.net/2011/08/08/gradle-keeping-libraries-up-to-date/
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -36,6 +36,12 @@ jar {
     }
 }
 
+run {
+    if (project.hasProperty("appArgs")) {
```
AsamK: Please add a comment here, how this can be used to pass arguments to signal-cli.

src/main/java/org/asamk/signal/Main.java
```diff
@@ -1293,6 +1296,59 @@ public void handleMessage(SignalServiceEnvelope envelope, SignalServiceContent c
...
+        public JsonDbusReceiveMessageHandler(Manager m, DBusConnection conn) {
+            super(m);
+            this.conn = conn;
+        }
+
+        @Override
+        public void handleMessage(SignalServiceEnvelope envelope, SignalServiceContent content, Throwable exception) {
+            super.handleMessage(envelope, content, exception);
+
+            if (envelope.isReceipt()) {
```
AsamK: Move this complete `if` to a separate function, and also use that function from `DbusReceiveMessageHandler`, so the code isn't just copied.

build.gradle
Riamse: okay

src/main/java/org/asamk/signal/Main.java
Riamse: I think a lot of the trouble is due to the fact that Java is single inheritance; in a multiple inheritance language like Python, I could just make a JsonDbusHandler that inherits from both the JsonHandler and the DbusHandler and just have two super().handleMessage() calls, which would've made life a lot easier.

I think it might actually be better to figure out if there's a better way to architect this part entirely, but I'm not sure if this method is the best way to do it. I think it will definitely be good in the long run to generalise it more than this.

Riamse: One thing I was considering was possibly something like this: you have an array of callbacks, and whenever you receive a message, each callback gets called with the message and does stuff to it. Then you can write a general json print callback, dbus-message-sending callback, etc... (The specifics will obviously vary, and there are drawbaks, but this is just one way to do it off the top of my head).

AsamK: yeah, that'll need more thinking.

PR #157, opened 2018-11-04
change SecureRandom Provider to best available (OS=`/dev/urandom`)

azet:
Currently `src/main/java/org/asamk/signal/util/Util.java` uses the
 `SHA1PRNG` algorithm  by Sun/Oracle to get randomness for `nextBytes` etc.

[_their_ (Oracle/Sun) documentation](https://docs.oracle.com/javase/8/docs/technotes/guides/security/SunProviders.html#SecureRandomImp) states: 
> ** On Solaris, Linux, and OS X, if the entropy gathering device in
> java.security is set to file:/dev/urandom or file:/dev/random, then
> NativePRNG is preferred to SHA1PRNG. Otherwise, SHA1PRNG is preferred.

and:

> The following algorithms are available in the SUN provider:
> 
> [...]
> 
> SecureRandom
> 
> SHA1PRNG (Initial seeding is currently done via a combination of system attributes and the java.security entropy gathering device)
> NativePRNG (nextBytes() uses /dev/urandom, generateSeed() uses /dev/random)
> NativePRNGBlocking (nextBytes() and generateSeed() use /dev/random)
> NativePRNGNonBlocking (nextBytes() and generateSeed() use /dev/urandom)

Using `/dev/urandom` (non-blocking, reworked in new linux kernel
releases) also reflects the opinion of many leading cryptographers over
any other solution of PRNGs out there, there you can find two crypto- and
a Java-specific blog post to underline this/my statement and for reassurance:
1. https://sockpuppet.org/blog/2014/02/25/safely-generate-random-numbers/
2. https://tersesystems.com/blog/2015/12/17/the-right-way-to-use-securerandom/
3. https://www.2uo.de/myths-about-urandom

Since we're not re-seeding a lot we're good using a blocking `/dev/random`
for seed generation (which thus also creates more secure seeds) and a faster, 
newly re-written implementation of `/dev/urandom` for (streaming) random bytes. 

I think this is the best choice we have available here.

I did  give a talk at SHA2017 on the subject matter, in case you're interested: 
https://www.youtube.com/watch?v=Yx4RroggBzI (including mentioned changes 
to Linux's CSPRNG design - though a lot of new improvements have since been
merged, especially since they're working on adding `Wireguard` to mainline).

### DIFF
--- a/src/main/java/org/asamk/signal/util/Util.java
+++ b/src/main/java/org/asamk/signal/util/Util.java
@@ -21,7 +21,7 @@ public class Util {
 
     private static SecureRandom getSecureRandom() {
         try {
-            return SecureRandom.getInstance("SHA1PRNG");
+            return SecureRandom.getInstance("NativePRNG");
         } catch (NoSuchAlgorithmException e) {
             throw new AssertionError(e);
         }


src/main/java/org/asamk/signal/util/Util.java
```diff
@@ -21,7 +21,7 @@ public static String getSecret(int size) {
 
     private static SecureRandom getSecureRandom() {
         try {
-            return SecureRandom.getInstance("SHA1PRNG");
+            return SecureRandom.getInstance("NativePRNG");
```
SalusaSecondus: May I recommend simply using `new SecureRandom()` instead? As per the documentation, on Solaris, Linux, and OS X, this is (currently) equivalent to `SecureRandom.getInstance("NativePRNG")`. However, on other platforms (such as Windows) you'll have problems as "NativePRNG" might not exist.  In all cases just using `new SecureRandom()` will give you what the JVM believes to be the best for your current system. This means that as the JVM is continually improved and developed, you will automatically get improvements in your random number generation.  (It also makes it easier for people to swap out implementations if they have other security requirements, such as running on a FIPS validated provider.)

A few other things to consider:
1. Immediately after creation, it's best for force seeding by calling `.nextBoolean()`. This keeps you safe if a consumer incorrectly calls `.setSeed()` as only the initial seeding has security concerns.
2. Entropy generation can be a thread-contention problem on many systems.  If you are worried about this, I recommend using a `ThreadLocal` construction.

Both of these things are how we do it in [some of our other projects](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/internal/Utils.java#L32).

azet: Good input. But as far as I can tell this project is (right now) not aimed at Windows systems at all and there doesn't seem to be a possibility to use it on Windows nor are there and users I've read of.

`ThreadLocal`: good input! You should discuss this with the original author.

azet: Wasn't aware, ok thanks, I was planning on chaning to `new SecureRandom` anyhow because of @SalusaSecondus comment.

azet: I'm wondering if there's a reason we should force seeding with `.nextBoolean()` now as I looked at the AWS code mentioned by @SalusaSecondus. Hm. I'm not sure if in this project there're any consumers that would be able to incorrectly call `.setSeed()` - it would imply the project itself being malicious. it's not a general purpose library after all.

AsamK:
Thanks for your PR.
I've made a commit to use the NativePRNG (if available) for generating all random numbers.

PR #168, opened 2018-11-20
SignalAccount Json File Fixes

up4:
Changes to make sure that the profile JSON file is loaded/saved correctly. I tested register, verify, addDevice, removeDevice, listDevices, setPin, removePin, send, receive.

Manager : removeLinkedDevices updates isMultiDevice and saves the account
Manager : addDevice, getLinkedDevices save the account

SignalAccount : save/load isMultiDevice
SignalAccount : save profileKey
SignalAccount : registrationLockPin doesn't automagically becomes the "null" string, and stays null if null


AsamK:
Thanks!

Issue #175, opened 2018-12-13
Can't send group messages.


taigrr:
`signal-cli -u +14xxxxxxxxx send -g X6zhFaxBxC99bK583Ub95w==  -m "test from cli"`
Results in:

> No recipients given
> Aborting sending.

Which is the same error when I don't specify a group at all. I think the check for a valid recipient happens without checking to see if the -g/--group parameter is set.

I'm completely new to this, and I haven't written java in years. Can someone else verify they have the same issue? 

`signal-cli -v`
> signal-cli 0.6.1

Herohtar:
I'm having the same issue. I can send messages to a specific user with no problem, but groups tell me that I haven't specified a recipient. Interestingly, if I give it a number AND a group it tells me _"You cannot specify recipients by phone number and groups a the same time"_ (sic) so at least part of the check seems to be working.

taigrr:
@Herohtar right, that's what happens for me as well. I'd bet that check happens after the recipient check.

Herohtar:
Well, I see the problem: https://github.com/AsamK/signal-cli/blob/7e897fa6d0aaa87646b51efa3ee1a5ecfaa3865e/src/main/java/org/asamk/signal/commands/SendCommand.java#L49-L53

This check happens before anything else when processing the `send` command, and the check for whether it's a group message is at the very end of that section.

The part of the code that checks for phone number and group at the same time happens elsewhere, so it's unrelated.

PR #176, opened 2018-12-14
Don't abort on empty recipient unless there was also no group specified.

Herohtar:
This fixes https://github.com/AsamK/signal-cli/issues/175.

The current recipient check prevents sending messages to a group. I've added a condition to the if statement so that it will continue sending the message if a group has been specified.

### DIFF
--- a/src/main/java/org/asamk/signal/commands/SendCommand.java
+++ b/src/main/java/org/asamk/signal/commands/SendCommand.java
@@ -52,7 +52,7 @@ public class SendCommand implements DbusCommand {
             return 1;
         }
 
-        if (ns.getBoolean("endsession") && ns.getList("recipient") != null && ns.getList("recipient").size() > 0) {
+        if (ns.getBoolean("endsession")) {
             try {
                 signal.sendEndSessionMessage(ns.<String>getList("recipient"));
                 return 0;
--- a/src/main/java/org/asamk/signal/commands/SendCommand.java
+++ b/src/main/java/org/asamk/signal/commands/SendCommand.java
@@ -46,6 +46,12 @@ public class SendCommand implements DbusCommand {
             return 1;
         }
 
+        if ((ns.getList("recipient") == null || ns.getList("recipient").size() == 0) && (ns.getBoolean("endsession") || ns.getString("group") == null)) {
+            System.err.println("No recipients given");
+            System.err.println("Aborting sending.");
+            return 1;
+        }
+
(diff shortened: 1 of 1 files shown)


Herohtar:
Just realized there might be issues with this if someone tries to send an `endsession` message to a group... might need some more work.

taigrr:
@Herohtar thanks.

While you're correct that the endsession message won't work, I'd argue that's a separate bug that shouldn't affect this issue.

I spent some time diving into the code and seeing how things work, noting the differences between the groupmessages and individual messages. Looks like the endsession messages just aren't implemented yet at all. It should be done, but I think it shouldn't hold back this PR so at least the group messaging functionality is available.

taigrr:
I manually built and applied @Herohtar 's fix and can confirm it works with group messages! @AsamK can you please merge this? Thanks!

src/main/java/org/asamk/signal/commands/SendCommand.java
```diff
@@ -46,7 +46,7 @@ public int handleCommand(final Namespace ns, final Signal signal) {
             return 1;
         }
 
-        if (ns.getList("recipient") == null || ns.getList("recipient").size() == 0) {
+        if ((ns.getList("recipient") == null || ns.getList("recipient").size() == 0) && ns.getString("group") == null) {
```
AsamK: Thanks for fixing this!
Can you please move the check for empty recipient into the ` if (ns.getBoolean("endsession")) {`, because that case doesn't support groups.
And check for recipient or group below that `if` block.

Herohtar: How's this? I don't like doing redundant checks, but I couldn't think of a better way to do it.

AsamK: Hm, this doesn't show an error if groupid and endsession are set and no recipients, but instead tries to send a group message in that case.

How about this check:
```

		        if ((ns.getList("recipient") == null || ns.getList("recipient").size() == 0) && (ns.getBoolean("endsession") || ns.getString("group") == null)) {
		            System.err.println("No recipients given");
		            System.err.println("Aborting sending.");
		            return 1;
		        }```

AsamK:
Nice, thanks!

taigrr:
Do you know how long it will take for this fix to make it into the AUR? I can run the version I compiled myself in the meantime, but I'd prefer to let aurman or yay take care of updates for me :)

AsamK:
As soon as I make a new release, which will be this weekend, as it's a rather bad bug.

PR #180, opened 2018-12-24
Add Support for MQTT brokers for publishing incoming Messages

peteh:
Hi,

I added an mqtt command that allows to set a broker as an optional command (otherwise it uses localhost) and publishes all incoming messages on the broker. The messages are published under the topic "signal-cli/messages/incoming/[usernamewithout+]/[data/typinginfo/receipt/]".

That's my first shot (actually second because i messaged up the commit history in my first pr) so I'd like to get some feedback if this is a good idea in general and also for the implementation.

Background: I want to integrate a signal bot into my snips voice framework (https://snips.ai/) to remotely check the state of light and turn it off from my phone if necessary. However that uses mqtt and runs on distributed systems (multiple raspberry pis in my flat) so dbus would probably work but I'd essentially build a dbus to mqtt wrapper. So I thought I could also add the broker directly to signal-cli.

Best regards, Pete

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,7 @@ dependencies {
     compile 'org.bouncycastle:bcprov-jdk15on:1.60'
     compile 'net.sourceforge.argparse4j:argparse4j:0.8.1'
     compile 'org.freedesktop.dbus:dbus-java:2.7.0'
+	compile 'org.eclipse.paho:org.eclipse.paho.client.mqttv3:1.0.2'
 }
 
 jar {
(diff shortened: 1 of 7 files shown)


streaps:
Could this PR merged? I think it's quite useful.

peteh:
> Could this PR be merged? I think it's quite useful.

I already continued working on it and also implemented sending messages through mqtt. You can check out the feature/mqtt-sending branch on my repository if you're interested. I currently use it to bridge signal into the snips framework, so i can ask it about the weather or turn my light on and off via text message in signal. 

However, I'd like to have somebody review it before merging.

PR #185, opened 2019-01-07
expose filename of attachments to json message handler

### LINKED ISSUE #184: Expose original filename in JSON output
- Filename is exposed in the plaintext log on stdout, but isn't available in the JSON output
- Proposed fix: expose the filename in the `org.asamk.signal.JsonAttachment` class

I will submit a PR with this fix momentarily!

thisisparker:
Fixes #184


AsamK:
Thanks!

PR #187, opened 2019-01-15
Add parameter to specify DBus name

user-invalid:
This way multiple instances of signal-cli can run in daemon-mode in
parallel. This was not possible previously as they tried to create the
same DBus.

This is rather my own quick fix and I guess there exists a cleaner way.
(Maybe let the parser parse the busname directly to `SIGNAL_BUSNAME`.)


AsamK:
I've implemented an alternative way to deal with multiple users: ca86c421ebb1a3930fa309a2d95a4090108c2e62
The `daemon` command can now be used without `-u +XXXXX` flag, in which case it will export all users in the data directory as separate dbus objects, instead of `/org/asamk/Signal` it will be exported as `/org/asamk/Signal/_XXXXX` for example.

PR #194, opened 2019-02-22
add the ability to pass in a message as JSON

vinnyt:
has a use case where I want to do something like `cat messge.json | signal-cli .. send <dest>` 

I was sending attachments in the JSON payload I pipe in vs adding -a for each one.


AsamK:
I've now implemented this with an API based on JSON-RPC, see #679

PR #206, opened 2019-06-05
Configure dbus name via commandline parameter

groupnet:
Same goal as [Add parameter to specify DBus name](https://github.com/AsamK/signal-cli/pull/187): configure dbus name via commandline parameter `--busname` to run multiple signal-cli daemons on different bus names in parallel.

The class [src/main/java/org/asamk/signal/DbusConfig.java](https://github.com/AsamK/signal-cli/compare/master...groupnet:feature/multipleDbus?expand=1#diff-06695f6173c02d6a507d9ec7fc94d69d) was adapted accordingly: the constant SIGNAL_BUSNAME was replaced by a variable and setter/getter functions. To keep the class uniform also the SIGNAL_OBJECTPATH was made variable but the _signalObjectPath_ was not exposed to the command line parameters.

Accidentally, I've also updated the Wiki's page (in the believe I was working on the fork), but removed the changes again. Anyway, you find the adaption here: [DBus service (c322d64)](https://github.com/AsamK/signal-cli/wiki/DBus-service/c322d64bc757a421cf68bd42e003f1f43fe776a4)

ToDo: adapt man page

### DIFF
--- a/data/org.asamk.Signal_NUMBER.conf
+++ b/data/org.asamk.Signal_NUMBER.conf
@@ -1,0 +1,16 @@
+<?xml version="1.0"?> <!--*-nxml-*-->
+<!DOCTYPE busconfig PUBLIC "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
+        "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
+
+<busconfig>
+        <policy user="signal-cli">
+                <allow own="org.asamk.Signal_%number%"/>
+                <allow send_destination="org.asamk.Signal_%number%"/>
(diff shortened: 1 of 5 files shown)


aguestuser:
curious: a few folks have made some attempt at PRs to paramaterize the dbus name so that multiple instances of signal-cli can run on the same machine in parallel. have any been merged?

AsamK:
I've implemented an alternative way to deal with multiple users: ca86c421ebb1a3930fa309a2d95a4090108c2e62
The `daemon` command can now be used without `-u +XXXXX` flag, in which case it will export all users in the data directory as separate dbus objects, instead of `/org/asamk/Signal` it will be exported as `/org/asamk/Signal/_XXXXX` for example.

Issue #211, opened 2019-07-19
signal-cli does not exit on authorization failure


groupnet:
**To reproduce:**
1. register number
2. backup config directory
3. unregister number
4. register number again
5. overwrite config directory with backup of step 2
6. try to send a message `signal-cli -u +... send "test"`

**Result:**
signal-cli halts (without exiting)
```bash
~ # signal-cli -u +... send "test"
Authorization failed, was the number registered elsewhere?
```

**Expected result:**
signal-cli exits with an error
```bash
~ # signal-cli -u +... send "test"
Authorization failed, was the number registered elsewhere?
~ # 
```

**Relevance:**
Not exiting is an issue especially when running signal-cli in daemon mode. As no error is thrown, the service is started and has the status "running" (as the dbus service is actually up and running). But trying to send a message via the dbus will fail without an error message.

PR #212, opened 2019-07-19
Throw error on failed authorization

groupnet:
- to exit signal-cli in case the number was registered elsewhere
 - solves issue [signal-cli does not exit on authorization failure](https://github.com/AsamK/signal-cli/issues/211) #211

### DIFF
--- a/src/main/java/org/asamk/signal/manager/Manager.java
+++ b/src/main/java/org/asamk/signal/manager/Manager.java
@@ -137,6 +137,7 @@ public class Manager implements Signal {
             }
         } catch (AuthorizationFailedException e) {
             System.err.println("Authorization failed, was the number registered elsewhere?");
+            throw e;
         }
     }
 


groupnet:
The error thrown in [Manager.java Line 140](https://github.com/AsamK/signal-cli/blob/f625309c03ecd53559be56f39a18066e60fb3d8a/src/main/java/org/asamk/signal/manager/Manager.java#L140) will be caught in [Main.java Line 94](https://github.com/AsamK/signal-cli/blob/f625309c03ecd53559be56f39a18066e60fb3d8a/src/main/java/org/asamk/signal/Main.java#L94)

With this pull request applied, the issue described in #211 will result in exiting signal-cli with error code 2 and the following output:
```bash
~ # signal-cli -u +... send "test"
Authorization failed, was he number registered elsewhere?
Error loading state file: Authorization failed!
~ # 
```

AsamK:
thanks

Issue #214, opened 2019-08-20
Unregister state is not saved on unregister


groupnet:
**To reproduce:**
1. register number
2. unregister number
3. try to register the same number again

**Result:**
signal-cli fails with message
```bash
~ # signal-cli -u +... register
Authorization failed, was the number registered elsewhere?
Error loading state file: Authorization failed!
```

**Expected result:**
signal-cli does not fail

**Work around:**
delete the data directory of the number in the config path (after unregister)
```bash
~ # rm -f [SIGNAL_CONFIG_PATH]/data/[PHONENUMBER]
~ # rm -rf [SIGNAL_CONFIG_PATH]/data/[PHONENUMBER].d
```

PR #215, opened 2019-08-20
Update SignalAccount storage on unregister

groupnet:
- save registered false state in the SignalAccount storage on unregister action
 - solves https://github.com/AsamK/signal-cli/issues/214


AsamK:
thanks

groupnet:
> thanks

You are very welcome. Keep up the great work! :+1:

PR #224, opened 2019-09-23
Added destination field of sync messages to JSON output

ra-rock:
Because the JSON output to stdout is not as robust as the human-readable output, I added the destination field from SentTranscriptMessage to JsonSyncMessage. This allows programs that read from stdout to better differentiate between messages. 

In another pull request, I will also look into creating a JsonReceiptMessage in order to add information from SignalServiceReceiptMessage to JsonMessageEnvelope, similarly to how it's handled in ReceiveMessageHandler.

### DIFF
--- a/src/main/java/org/asamk/signal/JsonSyncMessage.java
+++ b/src/main/java/org/asamk/signal/JsonSyncMessage.java
@@ -2,18 +2,24 @@ package org.asamk.signal;
 
 import org.whispersystems.signalservice.api.messages.multidevice.ReadMessage;
 import org.whispersystems.signalservice.api.messages.multidevice.SignalServiceSyncMessage;
+import org.whispersystems.signalservice.api.messages.multidevice.SentTranscriptMessage;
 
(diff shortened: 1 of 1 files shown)


AsamK:
Right, the destination number is missing from the sync sentMessage. But I think it should rather be inside the `sentMessage` object.
Like this:
```json
    "syncMessage": {
      "sentMessage": {
        "destination": "+DEST",
        "timestamp": 1569755652906,
        "message": "O",
        "expiresInSeconds": 0,
        "attachments": [],
        "groupInfo": null
      },
      "blockedNumbers": null,
      "readMessages": null
    },
```

Can you add a new class `JsonSyncDataMessage` that inherits from `JsonDataMessage` and has the `destination` field. Its constructor should take a `SentTranscriptMessage` and call its super constructor with the `sentTranscriptMessage.getMessage()`.

What do you think?

ra-rock:
Yeah that actually makes way more sense. On it.

Issue #174, opened 2018-12-10
Profile-Avatar?


zonk0101:
Hi!

is it possible to set a profile-avatar via CLI?

AsamK:
no, profiles aren't implemented in signal-cli yet.

mynameisflorian:
It would be very useful to be able to receive profiles from other users. Are there plans to add this feature in the near future?

AsamK:
Setting a profile avatar is implemented now with `updateProfile` command

technillogue:
I would also be interested in a way to receive profiles from other users.

Issue #98, opened 2017-09-06
profile support


Trolldemorted:
The official clients can now set/view profiles :tada: - it would be great if the cli master could do that too!

Atomic-Bean:
Setting is now possible, getting is an interesting one though. The result of getting the names would obviously be outputting it to stdout, but what exactly would getting the profile icon do? The individual user and group icons are already present in the config data so would the command just print the file path to stdout, or would it instead require an extra "path" variable from the user so signal-cli will copy the file to said path?

Tsjippy:
Just getting the name would be awesome

johnfreed:
I believe this can be closed, @AsamK .

Viewing arbitary profiles (at least the name, blocked status, and expiration timer) was implemented in the main program with the listContacts subcommand, and in the DBus daemon with the getContactName, IsContactBlocked, and listNumbers methods.

Setting these values for one's own profile can be done via the updateProfile subcommand and DBus method.

See the [Feature Matrix](https://github.com/AsamK/signal-cli/wiki/Feature-Matrix)

AsamK:
The `listContacts` command now outputs all profile information in json mode (5f941004f5006b286690df2a93bc47ace3b59270).

PR #227, opened 2019-09-24
Add commands to update profile name and avatar

lunar-debian:
These changes add two new commands: `setProfileName` and `setProfileAvatar` which allow to update the name and avatar visible by other users for the current profiles.

Maybe it would be better to have a single command `updateProfile` but it felt more straightforward that way for the quick hacking session I allowed myself to solve this.

Implements #174 and address part of #98.

### DIFF
--- a/src/main/java/org/asamk/signal/commands/SetProfileAvatarCommand.java
+++ b/src/main/java/org/asamk/signal/commands/SetProfileAvatarCommand.java
@@ -1,0 +1,41 @@
+package org.asamk.signal.commands;
+
+import java.io.IOException;
+import java.io.File;
+
+import net.sourceforge.argparse4j.impl.Arguments;
+import net.sourceforge.argparse4j.inf.Namespace;
+import net.sourceforge.argparse4j.inf.Subparser;
+import org.asamk.signal.manager.Manager;
+
+public class SetProfileAvatarCommand implements LocalCommand {
+
+    @Override
+    public void attachToSubparser(final Subparser subparser) {
+        subparser.addArgument("avatar")
+                .help("Path to new profile avatar");
+        subparser.help("Set the avatar for this profile");
+    }
+
+    @Override
+    public int handleCommand(final Namespace ns, final Manager m) {
+        if (!m.isRegistered()) {
+            System.err.println("User is not registered.");
+            return 1;
+        }
+
+        String avatarPath = ns.getString("avatar");
+	File avatarFile = new File(avatarPath);
+
+        try {
+            m.setProfileAvatar(avatarFile);
+	} catch (IOException e) {
+            System.err.println("UpdateAccount error: " + e.getMessage());
+            return 3;
+        }
+
+
+        return 0;
+    }
+
+}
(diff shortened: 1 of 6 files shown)


src/main/java/org/asamk/signal/manager/Manager.java
```diff
@@ -204,6 +205,14 @@ public void updateAccountAttributes() throws IOException {
         accountManager.setAccountAttributes(account.getSignalingKey(), account.getSignalProtocolStore().getLocalRegistrationId(), true, account.getRegistrationLockPin(), getSelfUnidentifiedAccessKey(), false);
     }
 
+    public void setProfileName(String name) throws IOException {
+        accountManager.setProfileName(account.getProfileKey(), name);
+    }
+
+    public void setProfileAvatar(File avatar) throws IOException {
+        accountManager.setProfileAvatar(account.getProfileKey(), Utils.createStreamDetailsFromFile(avatar));
```
AsamK: If I read the libsignal-service-java code correctly it doesn't close the input stream that gets passed here.
That means we need to store the streamDetails in a variable and call `streamDetails.getStream().close()` at the end of the method.

src/main/java/org/asamk/signal/commands/SetProfileAvatarCommand.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public int handleCommand(final Namespace ns, final Manager m) {
+        if (!m.isRegistered()) {
+            System.err.println("User is not registered.");
+            return 1;
+        }
+
+        String avatarPath = ns.getString("avatar");
+	File avatarFile = new File(avatarPath);
```
AsamK: replace the tab here with spaces

src/main/java/org/asamk/signal/commands/SetProfileNameCommand.java
```diff
@@ -0,0 +1,38 @@
...
+        if (!m.isRegistered()) {
+            System.err.println("User is not registered.");
+            return 1;
+        }
+
+        String name = ns.getString("name");
+
+        try {
+            m.setProfileName(name);
+	} catch (IOException e) {
```
AsamK: replace the tab here with spaces

src/main/java/org/asamk/signal/manager/Utils.java
```diff
@@ -56,6 +57,16 @@ static SignalServiceAttachmentStream createAttachment(File attachmentFile) throw
...
     }
 
+    static StreamDetails createStreamDetailsFromFile(File file) throws IOException {
+        InputStream stream = new FileInputStream(file);
+        final long size = file.length();
+        String mime = Files.probeContentType(file.toPath());
+        if (mime == null) {
+            mime = "application/octet-stream";
+        }
+	return new StreamDetails(stream, mime, size);
```
AsamK: replace the tab here with spaces

src/main/java/org/asamk/signal/manager/Manager.java
```diff
@@ -204,6 +205,14 @@ public void updateAccountAttributes() throws IOException {
...
+    }
+
```
AsamK: Please add a third method here `removeProfileAvatar`, that pass `null` as streamDetails to remove the avatar.

Issue #111, opened 2017-11-14
Add contacts


redshiftzero:
It would be great if a user could add a contact via the CLI (instead of [manually editing the config](https://github.com/AsamK/signal-cli/issues/12#issuecomment-220733780)). The reason being that currently in Signal Desktop there is no way to add the contact names for numbers. Ideally the contacts added via the CLI would then get synced to the associated devices. 

Related ticket: #12

RachelHendy:
This, pretty please :)

PR #228, opened 2019-09-24
Add commands to update contact names on Signal servers

lunar-debian:
As Signal Desktop does not allow its users to update the contact names, it would be nice to be able to update contact names from signal-cli when the latter is the master device.

The command `setContactName` allow to set the name for the given contact in the local store. `sendContacts` can then be used to push the updated contact list on Signal servers.

Address #111.

### DIFF
--- a/man/signal-cli.1.adoc
+++ b/man/signal-cli.1.adoc
@@ -208,6 +208,19 @@ number::
 	Specify the safety number or fingerprint of the key, only use this option if you have verified
 	the fingerprint.
 
+setContactName
+--------------
+Update name associated to a number on our contact list. This change is only local but can be synchronized to other devices by using `sendContacts` (see below).
+
+number::
+	Specify the contact phone number.
+
+name::
+	Specify the new name for this contact.
+
+sendContacts
+------------
+Update contact list on Signal servers.
 
 daemon
 ~~~~~~
--- a/src/main/java/org/asamk/signal/commands/SetContactNameCommand.java
+++ b/src/main/java/org/asamk/signal/commands/SetContactNameCommand.java
@@ -1,0 +1,34 @@
+package org.asamk.signal.commands;
+
+import net.sourceforge.argparse4j.impl.Arguments;
+import net.sourceforge.argparse4j.inf.Namespace;
+import net.sourceforge.argparse4j.inf.Subparser;
+import org.asamk.signal.manager.Manager;
+
+public class SetContactNameCommand implements LocalCommand {
+
+    @Override
+    public void attachToSubparser(final Subparser subparser) {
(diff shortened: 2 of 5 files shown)


man/signal-cli.1.adoc
```diff
@@ -208,6 +208,19 @@ number::
 	Specify the safety number or fingerprint of the key, only use this option if you have verified
 	the fingerprint.
 
+setContactName
+--------------
+Update name associated to a number on our contact list. This change is only local but can be synchronized to other devices by using `sendContacts` (see below).
+
+number::
```
AsamK: This should be uppercase `NUMBER` to match the format of the other commands.

src/main/java/org/asamk/signal/commands/SetContactNameCommand.java
```diff
@@ -0,0 +1,34 @@
+package org.asamk.signal.commands;
+
+import net.sourceforge.argparse4j.impl.Arguments;
+import net.sourceforge.argparse4j.inf.Namespace;
+import net.sourceforge.argparse4j.inf.Subparser;
+import org.asamk.signal.manager.Manager;
+
+public class SetContactNameCommand implements LocalCommand {
```
AsamK: I'd prefer this command to be more general, like `UpdateContactCommand`. That way it could be extended in the future to update other contact infos, like e.g. the contact's color.

I'd suggest to make the `name` argument a named parameter (`-n`, `--name`) but `.required(true)` for now.
So usage would be like this: `signal-cli ... updateContact +49XXX --name "FooBar"`

man/signal-cli.1.adoc
```diff
@@ -208,6 +208,19 @@ number::
...
+	Specify the contact phone number.
+
+name::
+	Specify the new name for this contact.
+
+sendContacts
+------------
+Update contact list on Signal servers.
```
AsamK: This description is not really correct. The Signal servers don't store the contacts list. They just forward the encrypted contacts list to the linked devices.

Suggestion:
`Send a synchronization message with the local contacts list to all linked devices. This command should only be used if this is the master device.`

Please also update the help text in `SendContactsCommand`.

man/signal-cli.1.adoc
```diff
@@ -208,6 +208,19 @@ number::
 	Specify the safety number or fingerprint of the key, only use this option if you have verified
 	the fingerprint.
 
+setContactName
+--------------
+Update name associated to a number on our contact list. This change is only local but can be synchronized to other devices by using `sendContacts` (see below).
```
AsamK: Please also add a note that the contact will be added if it doesn't exist yet.

PR #229, opened 2019-09-29
Added JsonSyncDataMessage class with destination field.

ra-rock:
Fixes discussed in PR #224


ra-rock:
Do you know why building might be failing? It builds locally perfectly fine.

Issue #239, opened 2019-11-02
Support sealed sender messages


xoxovm:
Hi,

with the version signal-cli 0.6.4 I get the following error message.

signal-cli -u + 49xxxxxxxxxx receive
Envelope from: (device: 0)
Timestamp: 1572726486758 (2019-11-02T20: 28: 06.758Z)
Sent by unidentified / sealed sender
Unknown message received

System: Arch Linux Linux 5.3.8-arch1-1

What is the problem here?

thanks

AsamK:
Looks like you managed to get a message from an unidentified sender. signal-cli doesn't support those completely yet. They should also be opt-in, so not sure how you got one.
Supporting them completely is on my TODO list.

xoxovm:
Everything works fine with version 0.6.3. As soon as an update to version 0.6.4 (pacman -u) is done, this error message appears.

xoxovm:
They are both numbers register, and still I get this error message.

kpcyrd:
I don't think this is Arch Linux specific, any way the title can be changed to "Sealed sender support"?

I have the same issue, I'm currently stuck writing a chat bot because the sender info is unavailable, sealed senders are enabled by default in signal 2.45.2.0 with no way to disable them.

kpcyrd:
I haven't tested this since I'm lacking a java development environment, but it seems instead of:
```
// envelope is SignalServiceEnvelope
envelope.getSourceAddress().getNumber()
```
you'd simply use:
```
// content is SignalServiceContent
content.getSender().getNumber()
```

AsamK:
Version 0.6.5 now supports receiving sealed sender messages

Issue #231, opened 2019-10-17
Read receipts and sealed sender


locksmyth:
Is there currently a way, or is there any plan, to implement read receipts and/or sealed sender.

AsamK:
Sending and receiving sealed sender messages is implemented now on master. I currently don't have any plans to implement sending read receipts.

PR #243, opened 2019-11-11
Add basic sealed sender support

kpcyrd:
This adds basic support to access the sender from messages sent by clients using sealed senders.

I'm still testing for the sender on the envelope first since this was necessary for some messages. I assume this is because signal-cli isn't sending with sealed-senders yet.

Related to #239 and #231.


AsamK:
thanks

julialen:
@AsamK did you already release sending with sealed-sender? If not, around when do you plan to release by?

AsamK:
yes receiving sealed-sender messages works with version 0.6.5

julialen:
Thanks for the response! My question was about sending with sealed sender, however -- does version 0.6.5 also include functionality for someone to send messages using sealed sender (or will a version come out soon with this functionality)?

AsamK:
Ah, no, sending sealed messages is not completely implemented yet.
I don't have an estimate, when it will be done.

PR #248, opened 2019-12-02
Add listContacts subcommand

visiblecode:
Add a listContacts subcommand that lists the currently registered contacts, similar to listDevices etc.

### DIFF
--- a/src/main/java/org/asamk/signal/commands/ListContactsCommand.java
+++ b/src/main/java/org/asamk/signal/commands/ListContactsCommand.java
@@ -1,0 +1,28 @@
+package org.asamk.signal.commands;
+
+import net.sourceforge.argparse4j.inf.Namespace;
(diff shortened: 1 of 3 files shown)


src/main/java/org/asamk/signal/commands/ListContactsCommand.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Override
+    public int handleCommand(final Namespace ns, final Manager m) {
+        if (!m.isRegistered()) {
+            System.err.println("User is not registered.");
+            return 1;
+        }
+        List<ContactInfo> contacts = m.getContacts();
+        for (ContactInfo c : contacts) {
+            System.out.println(String.format("Id: %s Name: %s", c.number, c.name));
```
AsamK: I think for contacts it would be better to use `Number:` instead of `Id:`.

Otherwise it looks good!

PR #254, opened 2019-12-29
[RFC] Handle sealed sender on receipt of data messages

visiblecode:
Since sealed sender became a thing, signal-cli fails to properly track the state of non-group threads as the envelope sender is always empty, so all thread state updates for such messages go to a thread with the id `""`.

This first shot at this is a quick hack that addresses the problem for regular messages. It doesn't yet address sync messages or other places that use the envelope sender.

Advice on style/how to structure this for real would be welcome.


AsamK:
I've pushed a commit with a similar change now, so closing this PR.

PR #256, opened 2019-12-30
Add support for uploading stickers.

signal-stickers:
Depends on
- https://github.com/Turasa/libsignal-service-java/pull/21

This adds a command to allow users to upload sticker packs:

`signal-cli -u <user> uploadStickerPack <zipfile>`
or
`signal-cli -u <user> uploadStickerPack path/to/manifest.json`

The zip file is expected to contain the following:

```
|_ manifest.json
|_sticker0.webp
|_sticker1.webp
|_...
|_stickerN.webp
```

The `manifest.json` is expected to contain the following:

```
{
  "title": "<mandatory>",
  "author": "<mandatory>",
  // Optional
  "cover": {
    "file": "<name of webp file, mandatory>",
    "emoji": "<optional>"
  },
  // Mandatory
  "stickers": [
    {
      "file": "<name of webp file, mandatory>",
      "emoji": "<optional>"
    }
    ...
  ]
  }
}
```

I have no issue writing a wiki entry.

### DIFF
--- a/src/main/java/org/asamk/signal/manager/Manager.java
+++ b/src/main/java/org/asamk/signal/manager/Manager.java
@@ -16,12 +16,9 @@
  */
 package org.asamk.signal.manager;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.asamk.Signal;
-import org.asamk.signal.AttachmentInvalidException;
-import org.asamk.signal.GroupNotFoundException;
-import org.asamk.signal.NotAGroupMemberException;
-import org.asamk.signal.TrustLevel;
-import org.asamk.signal.UserAlreadyExists;
+import org.asamk.signal.*;
 import org.asamk.signal.storage.SignalAccount;
 import org.asamk.signal.storage.contacts.ContactInfo;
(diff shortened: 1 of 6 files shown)


signal-stickers:
Updated to allow pointing to a manifest.json directly instead of having to use a zip file.

src/main/java/org/asamk/signal/manager/Manager.java
```diff
@@ -745,6 +725,138 @@ public void setExpirationTimer(String numberOrGroupId, int messageExpirationTime
         account.getThreadStore().updateThread(thread);
     }
 
+    public String uploadStickerPack(String path) throws IOException, StickerPackInvalidException {
```
AsamK: Nice work!

There's a bit too much logic for the sticker upload in this method for my taste.
I'd prefer to a have method in libsignal-service, that takes a data structure (similar to `SignalServiceStickerManifest` class, but with additional InputStream fields for the sticker image data).
So signal-cli would just parse the zip/files into that class and libsignal-service would handle the upload of the manifest and sticker files internally.
What do you think?

PR #260, opened 2020-01-19
Add support for blocking contacts and accounts

### LINKED ISSUE #115: blocked&expiry timer sync is coming to the official android clients
https://github.com/WhisperSystems/libsignal-service-java/commit/c8d7c3c00445a81b81e0a7305151cda4534ba299

It would be great if signal-cli would use it too

daniel-sch:
Implemented blocking (two new commands for blocking contacts and groups). The blocked status will be synced in the contact and group sync messages and with BlockedListMessages (I couldn't find any reference to BlockedListMessages in the code of Signal-Android and Signal-Desktop, so I couldn't test this part). This should close #115


src/main/java/org/asamk/signal/manager/Manager.java
```diff
@@ -686,6 +687,34 @@ public void setContactName(String number, String name) {
         account.save();
     }
 
+    @Override
+    public void setContactBlocked(String number, boolean blocked) {
+        ContactInfo contact = account.getContactStore().getContact(number);
```
AsamK: The number should be canonicalized before retrieving the contact, using `Utils.canonicalizeNumber(number, username)`

src/main/java/org/asamk/signal/manager/Manager.java
```diff
@@ -1175,6 +1206,26 @@ public void onMessage(SignalServiceEnvelope envelope) {
         }
     }
 
+    private boolean isMessageBlocked(SignalServiceEnvelope envelope, SignalServiceContent content) {
+        SignalServiceAddress source = envelope.getSourceAddress();
```
AsamK: With signal sealed messages the envelope doesn't always contain the actual source address.
This needs an additional check:
```java
            SignalServiceAddress source;
            if (!envelope.isUnidentifiedSender() && envelope.hasSource()) {
                source = envelope.getSourceAddress();
            } else if (content != null) {
                source = content.getSender();
            }

File: src/main/java/org/asamk/signal/manager/Manager.java [hunk 3]
```diff
@@ -1252,7 +1311,18 @@ private void handleMessage(SignalServiceEnvelope envelope, SignalServiceContent
                     }
                 }
                 if (syncMessage.getBlockedList().isPresent()) {
-                    // TODO store list of blocked numbers
+                    final BlockedListMessage blockedListMessage = syncMessage.getBlockedList().get();
+                    for (SignalServiceAddress address : blockedListMessage.getAddresses()) {
+                        if (address.getNumber().isPresent())
```
AsamK: please add braces `{}` for the if block

daniel-sch:
Thanks for your feedback, I changed the things you noted and also added the entries to the man page.

PR #289, opened 2020-04-01
Add SyncMessageReceived dbus push notifier

narodnik:
This is needed for when implementing custom chat. You need your own messages with timestamp managed by Signal so that messages appear in sync with each other.

This adds a notifier to the dbus similar to the existing one for received messages except instead of sender, it now has source and destination. Subscribing to this will tell you messages that you (the user) are sending through signal messenger.


AsamK:
thanks!

Issue #291, opened 2020-04-03
disappearing messages keep resetting


piramiday:
sometimes, somehow, my disappearing messages' setting gets reset.
this has happened in multiple chats, both between my accounts and with other folks, for no apparent reason.
I keep my `signal-cli` updated through many `receive` commands, so it's not the case that, for example, a "fresh" disappearing messages setting gets ignored by `signal-cli` upon sending a new message before having received the setting update.
I always send new messages after having received old updates.
any thoughts, @AsamK? thanks a lot for your awesome work! :+1:

piramiday:
just to be clear, I can see from my android device that: "signal-cli disabled disappearing messages".

piramiday:
further information: I have a chat between my primary phone number (signal for android + signal for desktop) and my secondary phone number (signal-cli).
using my primary account, I have set disappearing messages to expire (e.g. 1 week).

whenever such buggy behaviour occurs, I see:
- from my primary signal **desktop** instance, a message saying "_signal-cli disabled disappearing messages_"
- from my primary signal **android** instance, _no message_ related to disappearing messages,
- in **both** my primary signal android and desktop, I normally see the new message sent by signal-cli. indeed, it has been sent _without expiration timer_ !

I have never seen this issue if not with signal-cli.
any help is appreciated, since this is really an annoying bug. :cactus:

t351206:
I have the same issue!
After changing to "dissappearing message - 1 week" with the Android smartphone app I get the confirmation by the app. But immediately after signal_cli is sending a message it is deactivated again.

piramiday:
in my case, I am pretty sure that the message expiration time was reset after the `receive` command, not after any `send` command.

as a **workaround**, I've started to manually modify the database file (that is, `~/.local/share/signal-cli/data/+1234567890`) looking for the `messageExpirationTime` of the affected chats.
by putting this `sed` check right after my `signal-cli receive` command, I have not encountered other issues, so this seems to be a valid workaround.

still, a bugfix is unfortunately needed.

further information: after the disappearing messages are reset in such a way, I can see from my Android phone that the group chat still holds the _correct_ message disappearing time, therefore it's only signal-cli that sends messages with a _different_ (that is, infinite) disappearing time!

ThinkChaos:
I ran into this today, I can confirm the bug is triggered by a send as I never receive using signal-cli.

[...]

Issue #153, opened 2018-09-24
How to send disappearing message?


trananhtuan:
You mentioned this feature in v5.1 changelog but it's not documented anywhere.

PR #308, opened 2020-05-15
Contact config - message expiration time

matuskosut:
Could this be a way how to approach setting of expiration time for messages (on the contact level) through the client? As requested for example by https://github.com/AsamK/signal-cli/issues/153 and it may also help moving on with this one too https://github.com/AsamK/signal-cli/issues/291

If yes, I could add group configuration part or asses the current one if you have any feedback.

### DIFF
--- a/man/signal-cli.1.adoc
+++ b/man/signal-cli.1.adoc
@@ -254,7 +254,7 @@ Specify the new name for this contact.
 
 *-e*, *--expiration*::
 Set expiration time of messages (seconds).
-By default expiration is set to 0, no expiration.
+To disable expiration set expiration time to 0.
 
 === block
 
--- a/man/signal-cli.1.adoc
+++ b/man/signal-cli.1.adoc
@@ -252,6 +252,10 @@ Specify the contact phone number.
 *-n*, *--name*::
 Specify the new name for this contact.
 
+*-e*, *--expiration*::
+Set expiration time of messages (seconds).
+By default expiration is set to 0, no expiration.
+
 === block
 
 Block the given contacts or groups (no messages will be received).
--- a/src/main/java/org/asamk/signal/commands/UpdateContactCommand.java
+++ b/src/main/java/org/asamk/signal/commands/UpdateContactCommand.java
@@ -3,6 +3,8 @@ package org.asamk.signal.commands;
 import net.sourceforge.argparse4j.inf.Namespace;
 import net.sourceforge.argparse4j.inf.Subparser;
 
+import java.io.IOException;
+
 import org.asamk.signal.manager.Manager;
 import org.whispersystems.signalservice.api.util.InvalidNumberException;
 
@@ -41,6 +43,9 @@ public class UpdateContactCommand implements LocalCommand {
             }
         } catch (InvalidNumberException e) {
             System.out.println("Invalid contact number: " + e.getMessage());
+        } catch (IOException e) {
+            System.err.println("Update contact error: " + e.getMessage());
+            return 3;
         }
 
         return 0;
--- a/src/main/java/org/asamk/signal/commands/UpdateContactCommand.java
+++ b/src/main/java/org/asamk/signal/commands/UpdateContactCommand.java
@@ -18,7 +18,7 @@ public class UpdateContactCommand implements LocalCommand {
         subparser.addArgument("-e", "--expiration")
                 .required(false)
                 .type(int.class)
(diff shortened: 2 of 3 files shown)


src/main/java/org/asamk/signal/commands/UpdateContactCommand.java
```diff
@@ -15,6 +15,10 @@ public void attachToSubparser(final Subparser subparser) {
         subparser.addArgument("-n", "--name")
                 .required(true)
                 .help("New contact name");
+        subparser.addArgument("-e", "--expiration")
+                .required(false)
+                .type(int.class)
+                .help("Specify expiration time of messages");
```
AsamK, line 21: Please mention the unit here (seconds) and also document the argument in the `man/signal-cli.1.adoc` man page file.

src/main/java/org/asamk/signal/commands/UpdateContactCommand.java
```diff
@@ -30,6 +34,12 @@ public int handleCommand(final Namespace ns, final Manager m) {
 
         try {
             m.setContactName(number, name);
+
+            int expiration = 0;
```
AsamK, line 38: you can make `expiration` an `Integer` and already assign it here, like this:
```java
            Integer expiration = ns.getInt("expiration");
            if (expiration != null) {
                m.setExpirationTimer(number, expiration);
            }

File: src/main/java/org/asamk/signal/manager/Manager.java [hunk 3]
```diff
@@ -810,10 +810,12 @@ public String getGroupName(byte[] groupId) {
     /**
      * Change the expiration timer for a contact
      */
-    public void setExpirationTimer(SignalServiceAddress address, int messageExpirationTimer) {
```
AsamK, line 813: please keep the method with `SignalServiceAddress` parameter and just add a new method (with the same name) that takes a `String` and does the `canonicalizeAndResolveSignalServiceAddress` conversion.

src/main/java/org/asamk/signal/manager/Manager.java
```diff
@@ -810,10 +810,12 @@ public String getGroupName(byte[] groupId) {
...
-        ContactInfo c = account.getContactStore().getContact(address);
+    public void setExpirationTimer(String number, int messageExpirationTimer) throws InvalidNumberException {
+        ContactInfo c = account.getContactStore().getContact(canonicalizeAndResolveSignalServiceAddress(number));
         c.messageExpirationTime = messageExpirationTimer;
+        System.err.println("Set message expiration " + c.number + " -> " + c.messageExpirationTime + " seconds");
```
AsamK, line 816: I'd like to get rid of all console output in the Manager class, so please remove the println here as well.
As long as a command completes successfully, I'd prefer to not output anything as long as it doesn't provide additional information.

matuskosut: Sure this makes sense, removed it

src/main/java/org/asamk/signal/manager/Manager.java
matuskosut: edited, you can review the changes

matuskosut:
worked through the small things, good to review. When testing it sets the expiration correctly when first message after is sent, but I have to look a bit more on how to notify when it is set, and to make sure other devices get the change

man/signal-cli.1.adoc
```diff
@@ -252,6 +252,10 @@ Specify the contact phone number.
 *-n*, *--name*::
 Specify the new name for this contact.
 
+*-e*, *--expiration*::
+Set expiration time of messages (seconds).
+By default expiration is set to 0, no expiration.
```
AsamK, line 257: hm, this is a bit misleading, when adding a new contact, the expiration is set to 0 by default, but when updating a contact the expiration is left unchanged by default.

AsamK:
To notify the other contact, it should be enough to send a message with an expiration update:
```java
messageBuilder.withExpiration(contact.messageExpirationTime);
messageBuilder.asExpirationUpdate();
```
Sending it to the other contact will also trigger a sync message to the linked devices.

man/signal-cli.1.adoc
matuskosut: I changed the second part to `To disable expiration set expiration time to 0.`, does it feel better? Or would you remove that part completely?

matuskosut:
thanks for the hints, I added message sending

PR #317, opened 2020-06-02
Remove references to the native unix-java library

exquo:
Since version 0.6.8 signal-cli uses hypfvieh dbus-java, so installing the packages libunixsocket-java (Debian), libmatthew-java (Fedora) or libmatthew-unix-java (ArchLinux) is not necessary.


AsamK:
thanks

PR #341, opened 2020-08-25
update to jdk11

lkwg82:
updates to latest LTS to use newer language features


Trolldemorted:
This might break builds on phones running older java versions (?)

lkwg82:
What is the use case?

Trolldemorted:
You are right, as long as libsignal is unaffected this should be good to go

AsamK:
yeah, let's update it, even debian stable has java 11 now.
Thanks!

Issue #383, opened 2020-12-16
Receive with JSON is missing reactions field


Atomic-Bean:
If someone adds or removes a reaction, when you run `signal-cli -u +XXXXXXXXXX receive` then you'll see the emoji and its "Is remove" status, but if you run receive with `--json` enabled then there is no reaction information present.

Tested on latest signal-cli build in a 1-on-1 conversation.

AetherUnbound:
Additionally, any new group info for V2 groups does not appear in the JSON output (though it does appear in the non `--json` output)

Atomic-Bean:
Yeah I noticed that too, group stuff is just null as well as #151 (Which affects regular JSON mode, not just daemon).

Atomic-Bean:
It seems that the problem for this issue and a bunch of related ones lies in [this file](https://github.com/AsamK/signal-cli/blob/4d17e50323bcce21718d0bc2406c9f63db0f6f73/src/main/java/org/asamk/signal/json/JsonDataMessage.java#L20)

That is the class that takes Signal's `SignalServiceDataMessage` (The thing with all the message fields, reactions, quotes, etc included) and converts it to a format ready for the `JsonMessageEnvelope` [here](https://github.com/AsamK/signal-cli/blob/44851887894ed9162006ad9b12309e732aa3e876/src/main/java/org/asamk/signal/json/JsonMessageEnvelope.java#L38) which is then turned into a POJO [here](https://github.com/AsamK/signal-cli/blob/4d17e50323bcce21718d0bc2406c9f63db0f6f73/src/main/java/org/asamk/signal/JsonReceiveMessageHandler.java#L38) and is finally used in the receive output logic [here](https://github.com/AsamK/signal-cli/blob/master/src/main/java/org/asamk/signal/commands/ReceiveCommand.java#L165).

We can see its possible to get the same info in Json as the plain-text output because the plain-text output also uses `SignalServiceDataMessage` as seen [here](https://github.com/AsamK/signal-cli/blob/4d17e50323bcce21718d0bc2406c9f63db0f6f73/src/main/java/org/asamk/signal/ReceiveMessageHandler.java#L347)

[...]

Issue #151, opened 2018-09-04
Quoted replies do not include quotation information in JSON


eadmund:
Signal has had a quoted replies feature for awhile now. I've noticed that if there's a message which quotes another, `signal-cli` doesn't include the quotation in the JSON output of dæmon-mode.

matthew-piziak:
Yeah, this would be great. As it is right now, it is received as a standalone message, with no reference to the text being cited.

AsamK:
implemented in 58db3cbd53f3faec94ddfcd5e029865a380e6242

Atomic-Bean:
Commit 58db3cb only added it for regular output and not daemon mode as seen [here](https://github.com/AsamK/signal-cli/blob/master/src/main/java/org/asamk/signal/json/JsonDataMessage.java#L109). I believe PR #396 was supposed to add this for daemon, but it would be a bit out of date ATM. Same goes for #383.

heinrich-ulbricht:
I'm currently experimenting with DBus messaging in daemon mode from a .NET Core application. In the terminal I see quoted messages printed by the daemon (good), from the application I seem to be unable to access it (not so good).

The post from @Atomic-Bean indicates that the quote info is not accessible via DBus. Is this still the case?

Issue #388, opened 2020-12-21
Received v2 group info in JSON is missing name/members fields


Atomic-Bean:
First off, thanks for the quick update with v0.7.1. The features added are very useful 😄 

When I run the receive command in JSON mode I see something like this
```
{"envelope":{"source":"+XXXXXXXX","sourceDevice":2,"relay":null,"timestamp":1234567890,"dataMessage":{"timestamp":1234567890,"message":"SENDERS-MESSAGE-TEXT","expiresInSeconds":0,"attachments":[],"groupInfo":{"groupId":"[THE CORRECT GROUP ID]","members":null,"name":null,"type":"DELIVER"}},"syncMessage":null,"callMessage":null,"receiptMessage":null}}
```
The group Id is correct, but the name and members fields are missing. Running `listGroups -d` shows the full info for the group
```
Id: [THE CORRECT GROUP ID] Name: [THE CORRECT NAME]  Active: true Blocked: false Members: [+XXXXXXXXXX, +XXXXXXXXXX] Pending members: [] Requesting members: [] Link: -
```

Personally I don't think its worth bothering to print the members list here since for bigger groups that would just clutter the output and you can always use `listGroups -d` to see the full list (Also the plain-text print also doesn't show the members list either, which feels like a good choice). Either way its a low priority issue, the group ID is the important part and that's working perfectly

Using the latest build, v0.7.1 / 83d5d53 .

AsamK:
I've thought about it and won't add group name/members to the receive message output.
The information also be obtained from the listGroups command.

PR #389, opened 2020-12-22
Quotes, Mentions and Reactions in non-daemon JSON mode

Atomic-Bean:
This PR covers parts of #383 and #151

Features:
- Quotes, Mentions and Reactions will all now appear when running `signal-cli +XXXXXXXXX receive --json`. Their format is almost exactly the same as plain-text output mode (See more in Point 2). If the received message does not have some of these fields, then the default values are `null, [], null` respectively

Things that might be of concern:
1. In order to get the `targetAuthor` field of reactions to display the phone number, I had to pass down the `Manager` object by modifying multiple function signatures. The plain-text reaction output also used the manager for this information and I feel later on we'll need the manager down in the `JsonDataMessage` class for converting other UUIDs to phone numbers/names so I think this is fine.
2. I'm not sure what the intention with the `QuotedAttachment` plain-text output is. It outputs the filename and type from the `SignalServiceDataMessage.Quote.QuotedAttachment` object, then prints the `SignalServiceAttachment` object just like the regular attachment output. The only difference I saw was that the filename in the 2nd object was sometimes not present. So instead my JSON output code skips the first filename/type, makes a `JsonAttachment` object and overrides the filename if its not present.
3. In daemon mode all 3 parameters will always be null (Even Mentions, just to make it clear to the user it didn't receive an empty list). This is because I don't really know what's going on over there as I've never used daemons and I feel I'd break stuff for others if I messed with it.

### DIFF
--- a/src/main/java/org/asamk/signal/json/JsonGroupInfo.java
+++ b/src/main/java/org/asamk/signal/json/JsonGroupInfo.java
@@ -32,7 +32,6 @@ class JsonGroupInfo {
 
     JsonGroupInfo(SignalServiceGroupV2 groupInfo) {
         this.groupId = Base64.encodeBytes(GroupUtils.getGroupId(groupInfo.getMasterKey()));
-        // TODO populate members and name fields
         this.type = groupInfo.hasSignedGroupChange() ? "UPDATE" : "DELIVER";
     }
 
--- a/src/main/java/org/asamk/signal/json/JsonGroupInfo.java
+++ b/src/main/java/org/asamk/signal/json/JsonGroupInfo.java
@@ -32,6 +32,7 @@ class JsonGroupInfo {
 
     JsonGroupInfo(SignalServiceGroupV2 groupInfo) {
         this.groupId = Base64.encodeBytes(GroupUtils.getGroupId(groupInfo.getMasterKey()));
+        // TODO populate members and name fields
         this.type = groupInfo.hasSignedGroupChange() ? "UPDATE" : "DELIVER";
     }
 
--- a/src/main/java/org/asamk/signal/json/JsonMention.java
+++ b/src/main/java/org/asamk/signal/json/JsonMention.java
@@ -5,7 +5,7 @@ import org.whispersystems.signalservice.api.messages.SignalServiceDataMessage;
 
 public class JsonMention {
 
-    UUID uuid;    // If possible, it would be nice to resolve this into their phone-number/name. Same for plain-text output
+    UUID uuid;
     int start;
     int length;
 
--- a/src/main/java/org/asamk/signal/json/JsonMention.java
+++ b/src/main/java/org/asamk/signal/json/JsonMention.java
@@ -1,18 +1,22 @@
 package org.asamk.signal.json;
 
-import java.util.UUID;
+import org.asamk.signal.manager.Manager;
 import org.whispersystems.signalservice.api.messages.SignalServiceDataMessage;
+import org.whispersystems.signalservice.api.push.SignalServiceAddress;
 
 public class JsonMention {
 
-    UUID uuid;
+    String name;
     int start;
     int length;
 
-    JsonMention(SignalServiceDataMessage.Mention mention) {
-        this.uuid = mention.getUuid();
+    JsonMention(SignalServiceDataMessage.Mention mention, Manager m) {
+        this.name = m.resolveSignalServiceAddress(
+                new SignalServiceAddress(mention.getUuid(), null)
+        ).getLegacyIdentifier();
         this.start = mention.getStart();
         this.length = mention.getLength();
+
     }
 
 }
--- a/src/main/java/org/asamk/signal/json/JsonMention.java
+++ b/src/main/java/org/asamk/signal/json/JsonMention.java
@@ -1,0 +1,18 @@
+package org.asamk.signal.json;
+
+import java.util.UUID;
+import org.whispersystems.signalservice.api.messages.SignalServiceDataMessage;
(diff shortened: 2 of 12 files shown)


src/main/java/org/asamk/signal/json/JsonQuote.java
```diff
@@ -0,0 +1,41 @@
...
+
+        if (quote.getAttachments().size() > 0) {
+            this.attachments = new ArrayList<>(quote.getAttachments().size());
+
+            SignalServiceAttachmentPointer attachmentPointer;
+            for (SignalServiceDataMessage.Quote.QuotedAttachment quotedAttachment : quote.getAttachments()) {
+                JsonAttachment recentAttachment = new JsonAttachment(quotedAttachment.getThumbnail());
+
+                // Its possible the name might be missing, if it is then we'll use the other one
+                attachmentPointer = quotedAttachment.getThumbnail().asPointer();
```
AsamK, line 29: declare the variable here, it doesn't need to be preserved between iterations.
```suggestion
                SignalServiceAttachmentPointer attachmentPointer = quotedAttachment.getThumbnail().asPointer();
```

src/main/java/org/asamk/signal/json/JsonQuote.java
```diff
@@ -0,0 +1,41 @@
...
+        this.id = quote.getId();
+        this.author = quote.getAuthor().getLegacyIdentifier();
+        this.text = quote.getText();
+
+        if (quote.getAttachments().size() > 0) {
+            this.attachments = new ArrayList<>(quote.getAttachments().size());
+
+            SignalServiceAttachmentPointer attachmentPointer;
+            for (SignalServiceDataMessage.Quote.QuotedAttachment quotedAttachment : quote.getAttachments()) {
+                JsonAttachment recentAttachment = new JsonAttachment(quotedAttachment.getThumbnail());
```
AsamK, line 26: `getThumbnail()` can return `null`, if there is no thumbnail, e.g. for attached pdfs or other files.

AsamK (changes requested):
Thanks for this!
I think it's ok to pass the manager down to resolve the UUIDs.
The point about the `QuotedAttachment` is that the `SignalServiceAttachment` object is just an optional thumbnail for the actual attachment and can be null. The original attachment of the quoted message is not part of the quote.

src/main/java/org/asamk/signal/json/JsonQuote.java
Atomic-Bean: I see. I'm changing the `JsonQuote` class to use a new `JsonQuotedAttachment` object list that will correctly handle this

src/main/java/org/asamk/signal/json/JsonGroupInfo.java
```diff
@@ -32,6 +32,7 @@
 
     JsonGroupInfo(SignalServiceGroupV2 groupInfo) {
         this.groupId = Base64.encodeBytes(GroupUtils.getGroupId(groupInfo.getMasterKey()));
+        // TODO populate members and name fields
```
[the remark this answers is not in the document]
Atomic-Bean: Looking at the Signal-Android code for the `SignalServiceGroupV2` class, I'm not experienced enough to see how we'd get the name/members out of it. I'll remove the comment for now, and either way #388 exists so this can be pursued or "wont-fix" later.

src/main/java/org/asamk/signal/json/JsonMention.java
```diff
@@ -0,0 +1,18 @@
...
+import org.whispersystems.signalservice.api.messages.SignalServiceDataMessage;
+
+public class JsonMention {
+
+    UUID uuid;
+    int start;
+    int length;
+
+    JsonMention(SignalServiceDataMessage.Mention mention) {
+        this.uuid = mention.getUuid();
```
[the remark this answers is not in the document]
Atomic-Bean: That's a good idea. I noticed in multiple places (Mentions, quote's authors, some members from `listGroups` output, etc) it shows a UUID instead of a name. For Reactions we did this with by resolving the `SignalServiceAddress` with the manager and other things, but Mentions only [provides us](https://github.com/signalapp/Signal-Android/blob/9e7c55847e196165940ccb93c7bf55adbd3dac4a/libsignal/service/src/main/java/org/whispersystems/signalservice/api/messages/SignalServiceDataMessage.java#L558) the UUID.

I nearly gave up on adding this, but I finally found something that gets the mentioned person's phone number from the UUID. Do you want me to have both the UUID and phone number in the output or just the number? I'll also modify the plain-text outputs to show the phone number too.

Atomic-Bean:
New commit addresses all requested changes (I ended up removing the old UUIDs in the output and instead display the phone numbers for mentions/quotes. If you want me to display both let me know and I'll fix it). I also noticed quotes have their own "mentions" list so I added that for plain-text and JSON outputs.

I checked `listGroups` too, but it seems to be using the right method to get phone numbers. The UUIDs I see in my output belong to users in one of my groups, but I've never interacted with them or seen them post anything. I think that might be why I can't see their phone numbers.

AsamK (approved):
looks good!
A null check for `quote.getMentions()` is missing but I'll just add that myself :)

AsamK:
> I checked `listGroups` too, but it seems to be using the right method to get phone numbers. The UUIDs I see in my output belong to users in one of my groups, but I've never interacted with them or seen them post anything. I think that might be why I can't see their phone numbers.

Yeah, the new groups can be privacy preserving and don't disclose the members' phone numbers.

Issue #178, opened 2018-12-19
How to check if user exist?


thewalletbot:
What is the best way to check out if person has valid signal account if I know it;s phone numer.
For now I just send a message and I wait for an error. 
But it is quite stupid way I suppose. 
How can I make it in the most smart way?

AsamK:
There's currently no way with signal-cli.
There's a `getContact` method in `SignalServiceAccountManager` that can be used to check if a use is registered with the signal server.

Atomic-Bean:
I am currently implementing this. This is the format I have so far:
`signal-cli -u [YOUR-NUMBER] isRegistered [NUMBER [NUMBER ... ]] --json`

Any suggestions on the format or if I should rename the command?

thewalletbot:
> Any suggestions on the format or if I should rename the command?
looks OK for me.

PR #391, opened 2020-12-24
Command to check if number is registered

Atomic-Bean:
Implements #178 .

Command:
`signal-cli -u [YOUR-NUMBER] isRegistered [NUMBER [NUMBER ... ]] *--json`

Plain-text output looks like so:
[NUMBER]: [true / false]

JSON has String `number` and boolean `isRegistered` fields.

I'm not 100% on the command name, so feel free to suggest alternative names, command formats and other stuff.

### DIFF
--- a/src/main/java/org/asamk/signal/commands/Commands.java
+++ b/src/main/java/org/asamk/signal/commands/Commands.java
@@ -11,6 +11,7 @@ public class Commands {
         addCommand("addDevice", new AddDeviceCommand());
         addCommand("block", new BlockCommand());
         addCommand("daemon", new DaemonCommand());
+        addCommand("isRegistered", new IsRegisteredCommand());
         addCommand("link", new LinkCommand());
         addCommand("listContacts", new ListContactsCommand());
         addCommand("listDevices", new ListDevicesCommand());
--- a/src/main/java/org/asamk/signal/commands/Commands.java
+++ b/src/main/java/org/asamk/signal/commands/Commands.java
@@ -11,7 +11,7 @@ public class Commands {
         addCommand("addDevice", new AddDeviceCommand());
         addCommand("block", new BlockCommand());
         addCommand("daemon", new DaemonCommand());
-        addCommand("isRegistered", new IsRegisteredCommand());
+        addCommand("getUserStatus", new GetUserStatusCommand());
         addCommand("link", new LinkCommand());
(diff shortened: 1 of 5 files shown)


src/main/java/org/asamk/signal/commands/IsRegisteredCommand.java
```diff
@@ -0,0 +1,67 @@
...
+                }
+            } catch (IOException e) {
+                System.err.println(e.getMessage());
+            }
+        }
+
+        return 0;
+    }
+
+    private class JsonIsReceived {
```
AsamK, line 57:
```suggestion
    private class JsonIsRegistered {
```

src/main/java/org/asamk/signal/commands/Commands.java
```diff
@@ -11,6 +11,7 @@
         addCommand("addDevice", new AddDeviceCommand());
         addCommand("block", new BlockCommand());
         addCommand("daemon", new DaemonCommand());
+        addCommand("isRegistered", new IsRegisteredCommand());
```
AsamK, line 14: hm, maybe `getRegistrationStatus` or `getUserStatus` would be more flexible in case we add more information in the future.

AsamK (commented):
thanks! I added some ideas.

Atomic-Bean:
Sorry for the delayed response, I decided to take the rest of 2020 off.

1. Ah yes, I'll make is a JSON array
2. I didn't know of the plural function, but I will switch to that
3. Whoops, I used the receive command as a reference and made that mistake
4. I like `getUserStatus`. Like you said it will future proof it

src/main/java/org/asamk/signal/manager/Manager.java
```diff
@@ -345,6 +345,17 @@ public boolean isRegistered() {
...
 
+    /**
+     * This is used for checking if someone else's number is registered on Signal
+     *
+     * @param number The phone number in question
+     * @return True if registered, false otherwise
+     * @throws IOException if its unable to check if the user is registered
+     */
+    public boolean isUserRegistered(String number) throws IOException {
+        return this.accountManager.getContact(number).orNull() != null;
```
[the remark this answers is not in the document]
Atomic-Bean: Done. I don't fully understand `streams` / `Collectors.toMap()` so the loops/streams I used could probably be improved.

PR #396, opened 2020-12-30
Add a new command to read input from stdin instead of DBus

technillogue:
Hi, I was struggling to use the DBus interface for my project because it seems to depend on X11 and systemd to work correctly and noticed that some [other people](https://pypi.org/project/signal-cli-rest-api/) seemed to struggle with this as well. I was also frustrated that the JSON output included less information than regular output, and started addressing part of #383.


Atomic-Bean:
I'm glad you're able to add the missing JSON stuff for daemon mode :) . I didn't understand it enough to know how to do it. Also feel free to close issues #383 and #151 once this is all done and complete

technillogue:
darn, i've really screwed up the merges :/

technillogue:
@AsamK could you approve me for CI?

technillogue:
Alright, I think I finally got a reasonable diff and it compiles, I think I'm still missing some something causing ProtocolInvalidMessageException. It seems like this also needs to be fixed up with the new --output option.

The documentation I had had in my README should probably be put somewhere:

> This fork adds a stdio command, which is similar to the daemon --dbus command but read commands as JSON from stdin instead of exporting a DBus interface. It also includes reactions in JSON output.
> Currently, only sendMessage is available. The JSON interface is {"command": "sendMessage", "recipient": number, "content": message, "details": {"attachments": []}}. The details field is optional. As far as I can tell, attachments must be full absolute file paths.

The main thing it would be nice to add to this is sending messages to groups.

@maxcountryman @exquo if you were interested in contributing to this that would be lovely

exquo:
@technillogue thanks for pinging, ATM I can contribute some feedback:

- I would recommend splitting this into two logical PRs: stdin and reactions-JSON. 
	- Then, drop the latter :) (it has since been implemented, #383)
- A socket might be a better alternative to stdin, so that the daemon process can run in the background.

technillogue:
I think this should be good now?

i-infra:
Yay teamwork! Thanks for your time @AsamK @technillogue !!

i-infra:
@AsamK would you perchance review this for merge? I worked with @technillogue to refactor and improve logging, and I'm running this code in preproduction. Works great!

technillogue:
I have a different approach to this now that doesn't reimplement any commands

Issue #410, opened 2021-01-11
Download received stickers


Atomic-Bean:
While checking something else I noticed [this comment](https://github.com/AsamK/signal-cli/blob/master/src/main/java/org/asamk/signal/ReceiveMessageHandler.java#L424). I'm making this issue so that comment doesn't get lost.

Also seems the ["emoji" and "attachment" fields](https://github.com/signalapp/Signal-Android/blob/9e7c55847e196165940ccb93c7bf55adbd3dac4a/libsignal/service/src/main/java/org/whispersystems/signalservice/api/messages/SignalServiceDataMessage.java#L480) are missing from the receive command's output. When I output these, `emoji` appears to be a blank string? (Odd, and even so where would the emoji actually be on the UI? Its not the reaction and you can't seem to attach a message to a sticker post...) and `attachment` seems to be image information for the sticker itself which might be useful for downloading it.

(Also I'm currently working on a new PR that adds these two fields to the receive command as well as add Stickers for JSON mode, so you don't need to add those parts)

Atomic-Bean:
Ah, so if the sticker is sent from Signal Desktop then emoji equals `""`, but if its sent from something like the Android app then an emoji is present. Seems [every sticker has an emoji assigned to it](https://support.signal.org/hc/en-us/articles/360031836512-Stickers). Must be a desktop bug.

AsamK:
interesting ... I think the emoji in the sticker message is not strictly necessary. With the sticker pack id and key the complete sticker pack can be downloaded which contains all the stickers' emojis and also links to download all the sticker images.

So the question is if signal-cli should download the sticker attachment from the message or should just download the complete sticker pack. Also a `listStickerPacks` command could be added that lists all known sticker packs (already stored in the account file).
Though I'm not sure is if there's really a use case for stickers with signal-cli

Atomic-Bean:
Ah yeah, if something like `listStickerPacks` existed or even just reading signal-cli config info, then the emoji parameter in the receive command would be redundant. I guess also printing the attachment parameter is also redundant if later it can just be found with the key/id.

> Though I'm not sure is if there's really a use case for stickers with signal-cli

Stickers are weird. I think people's GUI programs that depend on signal-cli would find receiving/downloading sticker info useful and I think even normal signal-cli users would find sending stickers useful. But yeah imo its a low priority issue since stickers are very similar to attachments.

PR #418, opened 2021-01-13
Added sticker info to json output

Atomic-Bean:
Note that this doesn't actually implement #410 , instead this just adds the current 3 sticker fields to json output when running.

`signal-cli --output=json -u +61XXXXXXXX receive`


AsamK:
Thanks!

PR #459, opened 2021-01-21
doc(README): put links to manpage & wiki first

bentolor:
- Directly provide links to the complete man page below usage
- Extend explanation of `USERNAME`


AsamK:
Thanks, should make it easier to find the man page.

PR #482, opened 2021-01-31
Proposal for the Manager class to manage an ExecutorService for all SignalServiceMessageSender instances

Cutkh:
Hi,

first of all, thank you for this handy tool!

I propose that the Manager class manages an ExecutorService for all created instances of SignalServiceMessageSender instead of providing a null reference.

The background is the following:

I am currently building an MQTT-Signal gateway for my private home automation solution.
While testing possible implementations, I've found out that short-lived programs that directly use the Manager API and make calls to Manager#sendGroupMessage or similar 'send' methods prevent the program from exiting properly (without System.exit(); the behavior is nondeterministic, though).
The reason for this behavior is that SignalServiceMessageSender creates its own SingleThreadExecutor if it is not provided with an ExecutorService from the outside. This executor is only used for the single task of sending the message. The thereby created thread executes the task and then waits for the next tasks (which will never arrive). 
In short-lived programs, these executor services do not get garbage collected. Since the (non-daemon) thread is still waiting for a new task, the JVM does not stop, and the program hangs.  (See https://stackoverflow.com/questions/7728102/why-doesnt-this-thread-pool-get-garbage-collected for a similar behavior)

In the pull request, the Manager class manages a ThreadPoolExecutor that will be shut down once the Manager class's close-method is called. 
To keep it simple, I propose to use a default configuration that cannot be changed from the outside, which should be sufficient for almost all use cases: maximum of 10 thread; minimum of 0 threads, a thread will we shut down if it is idle for 10 milliseconds.

I believe that this pull request also improves the resource management for long-running clients in general because expensive to create threads are reused instead of being discarded after a single use and then block the computer until the garbage collector kicks in. 

Please let me know if you want me to make some further adjustments.

Thank you!

Dominik
P.s. sorry if this is more text then necessary, this is my first pull request ;-)

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/Manager.java
+++ b/lib/src/main/java/org/asamk/signal/manager/Manager.java
@@ -167,6 +167,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
@@ -188,6 +190,9 @@ public class Manager implements Closeable {
     private final GroupsV2Operations groupsV2Operations;
     private final SignalServiceMessageReceiver messageReceiver;
     private final ClientZkProfileOperations clientZkProfileOperations;
+
+    private final ExecutorService executor = new ThreadPoolExecutor(0, 10, 10, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>()); //Todo added by me
+
 
     private SignalServiceMessagePipe messagePipe = null;
     private SignalServiceMessagePipe unidentifiedMessagePipe = null;
@@ -526,7 +531,6 @@ public class Manager implements Closeable {
     }
 
     private SignalServiceMessageSender createMessageSender() {
-        final ExecutorService executor = null;
         return new SignalServiceMessageSender(serviceEnvironmentConfig.getSignalServiceConfiguration(),
                 account.getUuid(),
                 account.getUsername(),
@@ -2533,6 +2537,8 @@ public class Manager implements Closeable {
     }
 
     void close(boolean closeAccount) throws IOException {
+        executor.shutdown();
+
         if (messagePipe != null) {
             messagePipe.shutdown();
             messagePipe = null;
--- a/lib/src/main/java/org/asamk/signal/manager/Manager.java
+++ b/lib/src/main/java/org/asamk/signal/manager/Manager.java
@@ -167,8 +167,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


lib/src/main/java/org/asamk/signal/manager/Manager.java
```diff
@@ -189,6 +191,9 @@
     private final SignalServiceMessageReceiver messageReceiver;
     private final ClientZkProfileOperations clientZkProfileOperations;
 
+    private final ExecutorService executor = new ThreadPoolExecutor(0, 10, 10, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>());
+
+
```
AsamK, line 196: a single empty line is enough here.

lib/src/main/java/org/asamk/signal/manager/Manager.java
```diff
@@ -189,6 +191,9 @@
     private final SignalServiceMessageReceiver messageReceiver;
     private final ClientZkProfileOperations clientZkProfileOperations;
 
+    private final ExecutorService executor = new ThreadPoolExecutor(0, 10, 10, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>());
```
AsamK, line 194: Initializing the ThreadPoolExecutor with a LinkedBlockingQueue effectively makes this a single thread executor.
See javadoc:
>  Unbounded queues. Using an unbounded queue (for example a LinkedBlockingQueue without a predefined capacity) will cause new tasks to wait in the queue when all corePoolSize threads are busy. Thus, no more than corePoolSize threads will ever be created. (And the value of the maximumPoolSize therefore doesn't have any effect.) This may be appropriate when each task is completely independent of others, so tasks cannot affect each others execution; for example, in a web page server. While this style of queuing can be useful in smoothing out transient bursts of requests, it admits the possibility of unbounded work queue growth when commands continue to arrive on average faster than they can be processed.

I'm fine with that, but then you can just use `Executors.newSingleThreadExecutor()` :)
Also using a coreSize of 1 is probably a good idea (which newSingleThreadExecutor already does), the thread will be started lazily anyway.

AsamK (changes requested):
Thanks for this, the long explanation is appreciated :)

lib/src/main/java/org/asamk/signal/manager/Manager.java
Cutkh: Ups, I've trusted the answer of StackOverflow and just skimmed the Javadoc afterward - classic mistake ^^'

I think a single thread might be a bottleneck in scenarios where one wants to send multiple messages on a mediocre internet connection. Having read the Javadoc around ThreadPoolExecutor and Executors more carefully, I suggest using a `Executors.newCachedThreadPool()`. Since the number of threads is virtually unbounded, the behavior is almost similar to before the change - with the benefit that threads are reused and that the thread pool is shutdown on closing.

lib/src/main/java/org/asamk/signal/manager/Manager.java
```diff
@@ -189,6 +190,8 @@
     private final SignalServiceMessageReceiver messageReceiver;
     private final ClientZkProfileOperations clientZkProfileOperations;
 
+    private final ExecutorService executor = Executors.newCachedThreadPool();
+
```
Cutkh, line 194: I think a single thread (`newSingleThreadExecutor()`) might be a bottleneck in scenarios where one wants to send multiple messages on a mediocre internet connection. Having read the Javadoc around Executors more carefully, I suggest using a `Executors.newCachedThreadPool()`. That way, the behavior is almost similar to before the change - with the benefit that threads are reused and that the thread pool is shutdown on closing.

AsamK (approved):

PR #506, opened 2021-02-08
fix for gradle >=6.8

aanno:
Dear AsamK,

I had small problems building signal-cli with gradle >=6.8 and fixed them. Perhaps you can consider this PR for inclusion.

Kind regards,

aanno

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -13,7 +13,7 @@ java {
 }
 
 application {
-    mainClass.set("org.asamk.signal.Main")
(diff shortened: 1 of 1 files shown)


AsamK:
I'm building successfully with gradle 6.8 and mainClassName is marked as deprecated ...
Which error did you get?

aanno:
Bummer! I thought that I'm using gradle 6.8.2 _but_ it was gradle 6.2.2. Hence this PR is invalid. Very sorry for the noise!

PR #522, opened 2021-02-15
Add updateProfile to Dbus

bublath:
Profile the updateProfile to the Dbus interface.
Is the exception handling ok?

### DIFF
--- a/src/main/java/org/asamk/Signal.java
+++ b/src/main/java/org/asamk/Signal.java
@@ -41,10 +41,6 @@ public interface Signal extends DBusInterface {
 
     List<byte[]> getGroupIds();
 
-    List<String> listGroups(); 
-
-    List<String> listContacts();
-    
     String getGroupName(byte[] groupId);
 
     List<String> getGroupMembers(byte[] groupId);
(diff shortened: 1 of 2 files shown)


AsamK:
Thanks, the profile changes look good. Just the code formatting is a bit off (missing spaces and some tabs for indentation).
If you use IntellIJ you can use the auto formatter to format the code.

I don't like the `listGroups` and `listContacts` commands much, there's already a `getGroupIds` method, so maybe something similar for contacts? Though I'm no expert on dbus API design ...

bublath:
Another option for groups and contacts would be to split that up:
getContacts(): Just returns an array of numbers
getContactNumber(string name): Query if there is a contact where you only have the name
isContactBlocked(string contact):
isGroupBlocked(byte[] groupID);
isGroupActive(byte[] groupID);

That would also enable me to get the information I need - but with more calls...

What do you think?

bublath:
Let's take this one step at a time. I reverted the listGroups and listContacts and fixed the tab formatting for the profile change.

My problem statement for listGroups and listContacts is to enable users to
- use contact names instead of numbers without needing to cache them in my application (avoid double data storage)
- know if contacts or groups are blocked or inactive to handle this (warning, display appropriately)
- know what contacts are there in the first place, so users can choose

I had started a thread in the issues that actually is about the best way of implementing this. We can move the discussion there if this pull request gets approved: https://github.com/AsamK/signal-cli/issues/524

Issue #512, opened 2021-02-12
Extract and display received contacts from DataMessage


Atomic-Bean:
Signal's Data Message can contain [a list of contacts](https://github.com/signalapp/Signal-Android/blob/master/libsignal/service/src/main/java/org/whispersystems/signalservice/api/messages/SignalServiceDataMessage.java#L34). ATM this is not outputted in the receive command.

I'll probably make a PR for this eventually.

Atomic-Bean:
So it seems there's two parts to this. The first is adding all the fields from the contacts to our output and the second is downloading the contact's avatar, if there is one (Which uses the same format as a normal attachment).

When we receive a contact/s message, I'm currently downloading any avatars to the attachments folder, but I think we'd probably want to do the Avatars folder instead. Where do you want it?

AsamK:
I think storing it in the attachments folder is fine. The shared contacts are not persisted, so storing it in the avatars folder doesn't feel right.

PR #529, opened 2021-02-17
Output "SharedContacts" field from a SignalDataMessage

Atomic-Bean:
This PR:
- Implements #512 
- Outputs the full SharedContact details when we receive a message with this field present (For both plaintext and json)
- Also saves the avatar, if present, to the `attachments/` directory

One question I have. There are lots of the fields within the SharedContact and most of the time a lot of them aren't present/null. For Json mode, should I add Json Properties to omit the field if null? I think they should stay for compatibility, but I'm not sure.


src/main/java/org/asamk/signal/ReceiveMessageHandler.java
```diff
@@ -421,16 +421,119 @@ private void handleSignalServiceDataMessage(SignalServiceDataMessage message) {
...
+                        if (email.getValue() != null) {
+                            System.out.println("     - Value: " + email.getValue());
+                        }
+                        if (email.getType() != null) {
+                            System.out.println("     - Type: " + email.getType());
+                        }
+                        if (email.getLabel().isPresent() && !email.getLabel().get().isBlank()) {
+                            System.out.println("     - Label: " + email.getLabel().get());
+                        }
+                    }
```
AsamK, line 496: missing `i++`
though I'm not sure if numbering the lists is very useful anyway ...

src/main/java/org/asamk/signal/json/JsonContactAddress.java
```diff
@@ -0,0 +1,59 @@
...
+        pobox = getValueIfActuallyPopulated(address.getPobox().orNull());
+        neighborhood = getValueIfActuallyPopulated(address.getNeighborhood().orNull());
+        city = getValueIfActuallyPopulated(address.getCity().orNull());
+        region = getValueIfActuallyPopulated(address.getRegion().orNull());
+        postcode = getValueIfActuallyPopulated(address.getPostcode().orNull());
+        country = getValueIfActuallyPopulated(address.getCountry().orNull());
+        if (country == null) {
+            System.out.println("Is null");
+        } else {
+            System.out.println("Present: " + country);
```
AsamK, line 56: remove output here

src/main/java/org/asamk/signal/json/JsonContactEmail.java
```diff
@@ -0,0 +1,30 @@
...
+    @JsonProperty
+    private final String value;
+
+    @JsonProperty
+    private final SharedContact.Email.Type type;
+
+    @JsonProperty
+    private final String label;
+
+    private String getValueIfPresent(String string) {
```
AsamK, line 18: Maybe change this method to take an `Optional<String>`?
Also you can move it as a static method to `util/Util.java` and reuse it.

AsamK (changes requested):
Thanks, I just have a few comments.
I'd say it's fine to output the `null` values here, I mostly omitted null fields for mutually exclusive fields or for very rarely used fields.

src/main/java/org/asamk/signal/ReceiveMessageHandler.java
Atomic-Bean: Thanks. The idea behind it is the `"  - No. " + i` line is used to separate the different Phones/Emails/Addresses otherwise you get output like this
```
 - Email details:
    - Value: x1
    - Type: y1
    - Label: z1
    - Value: x2
    - Type: y2
    - Label: z2
```
Maybe I can change the `"  - No. " + i` to instead be "  - Email: " or some other message.

src/main/java/org/asamk/signal/json/JsonContactEmail.java
Atomic-Bean: All good. For some reason I thought you didn't want Optionals in the codebase, but I'll fix that now.

AsamK: Yeah, I'm not too fond of Optionals, but for the helper method it's totally fine.

PR #543, opened 2021-02-25
Dbus enhancements

bublath:
I started on significantly extending the functional coverage via Dbus.
This pull request is not yet intended for merge (e.g. I know I need to get rid of the tabs :) ).
I mainly wanted to get feedback about the direction.
1. I changed one of the "sendMessage" methods - more details in discussion https://github.com/AsamK/signal-cli/discussions/542
2. I'm "merging" contacts and profiles to get better coverage of names (signal-cli accounts often have no contacts)
3. I'm fetching a combined list of numbers based on identities and contacts to have a more complete list
4. I still have doubts about a simple "listNumbers" method and various methods to get details like "isBlocked", "getContactName" compared to a method that will return a list that already contains numbers, names and status
5. I made a manager method public since I had good use, but maybe there is a better way

Thanks in advance for the feedback.

### DIFF
--- a/src/main/java/org/asamk/signal/dbus/DbusSignalImpl.java
+++ b/src/main/java/org/asamk/signal/dbus/DbusSignalImpl.java
@@ -43,7 +43,7 @@ public class DbusSignalImpl implements Signal {
     }
 
     @Override
-    public long sendMessageSingle(final String message, final List<String> attachments, final String recipient) {
+    public long sendMessage(final String message, final List<String> attachments, final String recipient) {
         var recipients = new ArrayList<String>(1);
         recipients.add(recipient);
         return sendMessage(message, attachments, recipients);
@@ -65,9 +65,8 @@ public class DbusSignalImpl implements Signal {
         }
     }
 
-    private static void checkSendMessageResults(
-            long timestamp, List<SendMessageResult> results
-    ) throws DBusExecutionException {
+    private static void checkSendMessageResults(long timestamp, List<SendMessageResult> results)
+            throws DBusExecutionException {
         if (results.size() == 1) {
             checkSendMessageResult(timestamp, results.get(0));
             return;
@@ -90,14 +89,6 @@ public class DbusSignalImpl implements Signal {
 
     @Override
     public long sendMessage(final String message, final List<String> attachments, final List<String> recipients) {
-		System.out.println("Sending messsage:"+message);
-		System.out.println("Sending rec:"+recipients);
-		for (String r : recipients) {
-			System.out.println("to:"+r);
-		}
-		for (String r : attachments) {
-			System.out.println("at:"+r);
-		}
         try {
             final var results = m.sendMessage(message, attachments, recipients);
             checkSendMessageResults(results.first(), results.second());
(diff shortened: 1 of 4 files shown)


bublath:
I switched my editor to VSCode/IntelliJ - hope formatting works now.
After some consideration I reverted the sendMessage change. I think I found a way to use a different package in Perl that handles the difference in arguments correctly.
Completed isContactBlocked(). Currently I throw an exception - could also just return "false" since a non-existing contact is not blocked. Would make things easier for the client.

Any feedback?

src/main/java/org/asamk/signal/dbus/DbusSignalImpl.java
```diff
@@ -248,24 +247,134 @@ public boolean isRegistered() {
...
+    @Override
+    public String version() {
+        return BaseConfig.PROJECT_VERSION;
+    }
+
+    // Create a unique list of Numbers from Identities and Contacts to really get
+    // all numbers the system knows
+    @Override
+    public List<String> listNumbers() {
+        List<String> numbers = new ArrayList<>();
```
AsamK, line 275: this method could be written more concisely with streams:
```java
return Stream.concat(m.getIdentities().stream().map(i -> i.getAddress().getNumber().orNull()),
                m.getContacts().stream().map(c -> c.number))
                .filter(Objects::nonNull)
                .distinct()
                .collect(Collectors.toList())
```

src/main/java/org/asamk/signal/dbus/DbusSignalImpl.java
```diff
@@ -140,13 +143,14 @@ public long sendGroupMessage(final String message, final List<String> attachment
...
+    // Profile names separate firstname/lastname by a \0 - replace by space
+    // otherwise the interface will hickup
     @Override
     public String getContactName(final String number) {
-        try {
-            return m.getContactName(number);
-        } catch (InvalidNumberException e) {
-            throw new Error.InvalidNumber(e.getMessage());
-        }
+        String name = m.getContactOrProfileName(number).replace("\0", " ");
```
AsamK, line 152: Interesting, but this shouldn't be in the dbus layer. I've added a `getDisplayName()` method to the profile class, so you can remove the replacing here.

src/main/java/org/asamk/signal/dbus/DbusSignalImpl.java
```diff
@@ -248,24 +247,134 @@ public boolean isRegistered() {
...
+            if (numbers.indexOf(c.number) == -1) {
+                numbers.add(c.number);
+            }
+        }
+
+        return numbers;
+    }
+
+    @Override
+    public String getContactNumber(final String name) {
```
AsamK, line 297: This method is quite nondeterministic in case there are multiple profiles with the same name. Which is quite common as many people just use the first name.
How about making this method return a list of strings, so in case of duplicates multiple numbers can be returned?

src/main/java/org/asamk/signal/dbus/DbusSignalImpl.java
```diff
@@ -248,24 +247,134 @@ public boolean isRegistered() {
...
+                return c.number;
+            }
+        }
+        // Try profiles if no contact name was found
+        for (IdentityInfo identity : m.getIdentities()) {
+            String number = identity.getAddress().getNumber().orNull();
+            if (number != null) {
+                var address = Utils.getSignalServiceAddressFromIdentifier(number);
+                var profile = m.getRecipientProfile(address);
+                String profileName = profile.getName().replace("\0", " ");
```
AsamK, line 311:
```suggestion
                String profileName = profile.getDisplayName().replace("\0", " ");
```

AsamK (commented):
Looks good.
I think isContactBlocked can just return false.

bublath:
I basically included all of your proposals, though fixing 2 small issues and added a trim() to the DisplayName, since as you pointed out people could just use the firstname, but then the replace would result to "name " (trailing space). trim() should fix that.
P.S. I was really impressed about your "stream" solution. Took me a moment to grasp it :)

PR #556, opened 2021-03-03
some small dbus return code fixes

bublath:
Need to distinguish empty profile name (return "") vs number not found (null) . Dbus does not really like "null" and it would create an exception.
Handling of wrong link uri was also changed to trigger a correct exception for the same reason.


lib/src/main/java/org/asamk/signal/manager/Manager.java
```diff
@@ -2365,9 +2365,9 @@ public String getContactOrProfileName(String number) {
 
         final var profileEntry = account.getProfileStore().getProfileEntry(address);
         if (profileEntry != null && profileEntry.getProfile() != null) {
-            return profileEntry.getProfile().getDisplayName();
+            String name = profileEntry.getProfile().getDisplayName();
+            return name==null?"":name;
```
AsamK, line 2369: This null check should just be in the dbus layer.

bublath: On the Dbus layer I can no longer distiguish if the null is due to emtpy name in profile or number not found/badly formatted.
Alternative would be to return null for empty and throw an exception for "not found".

bublath: To iterate a bit more maybe on the case: My client program can display the list of contacts or group members. Here an empty profile name (which can happen e.g. if the account was created by signal-cli) should simply fall back to the phone number (like your writer that uses the same method).
However the client also tries to translate numbers to names and it can happen that this is done for an unknown or badly formatted number. In that case I would rather not return an empty string, but raise an error to the user.

man/signal-cli-dbus.3.adoc
```diff
@@ -0,0 +1,205 @@
+/////
+vim:set ts=4 sw=4 tw=82 noet:
+/////
+:quotes.~:
+
+= signal-cli-dbus (3)
```
AsamK, line 6: I think 3 is not the best place for it, it's for programming libraries.
Most dbus descriptions on my system are in 5.

man/signal-cli-dbus.3.adoc
```diff
@@ -0,0 +1,205 @@
...
+
+== Name
+
+DBus API for signal-cli - A commandline and dbus interface for the Signal messenger
+
+== Synopsis
+
+*signal-cli* [--config CONFIG] daemon --dbus-system
+
+*dbus-send* --system --type=method_call --print-reply --dest="org.asamk.Signal" /org/asamk/Signal org.asamk.Signal<method> [string:<string argument>] [array:<type>:<array argument>]
```
AsamK, line 16:
```suggestion
*dbus-send* --system --type=method_call --print-reply --dest="org.asamk.Signal" /org/asamk/Signal org.asamk.Signal.<method> [string:<string argument>] [array:<type>:<array argument>]
```

man/signal-cli-dbus.3.adoc
```diff
@@ -0,0 +1,205 @@
...
+
+See signal-cli (1) for details on the application.
+
+This documentation handles the supported methods when running signal-cli as a DBus daemon.
+
+The method are described as follows:
+
+Return<type> = Method( arg1<type>, arg2<type>, ...)
+
+Where type are according to DBus specification:
```
AsamK, line 29:
```suggestion
Where `type` is according to DBus specification:
```

man/signal-cli-dbus.3.adoc
```diff
@@ -0,0 +1,205 @@
...
+
+== Description
+
+See signal-cli (1) for details on the application.
+
+This documentation handles the supported methods when running signal-cli as a DBus daemon.
+
+The method are described as follows:
+
+Return<type> = Method( arg1<type>, arg2<type>, ...)
```
AsamK, line 27: stylistic remark, please make sure all methods use the same formatting
- camel case for names (no spaces in argument name)
- no spaces before or after brackets
- a single space after each comma
- maybe put the return type behind the method?

```suggestion
Method(arg1<type>, arg2<type>, ...) → Return<type>
```

man/signal-cli-dbus.3.adoc
```diff
@@ -0,0 +1,205 @@
...
+version<s> version()::
+* version : Version string of signal-cli
+
+result<b> isRegistred::
+* result : Currently always returns 1=true
+
+== Signals
+
+ SyncMessageReceived (timestamp<x>,sender<s>,destination<s>,groupId<ay>,message<s>,attachments<as>)::
+Currently not used
```
AsamK, line 184: the sync message is received when the user sends a message from a linked device

AsamK (commented):
Thanks!
In the future please create separate PRs for different changes, makes it easier to review.

bublath:
Sorry for the confusion with the two commits. When I removed the man page I forgot to add the changed sources.
I hope now its ok. getDisplayName now returns "" and getContactOrProfile name is untouched.
I will work a bit more on the man page and resubmit it as (5)

AsamK (approved):

PR #563, opened 2021-03-07
Fix README formatting

0mp:
By adding two spaces at the end of a line, we force the Markdown renderer to insert a new line.


AsamK:
Thanks for noticing.
Though I don't like trailing white spaces (and my editor might remove them anytime :sweat_smile: ).
I've added a new line instead, looks a bit different but I think that's fine.

0mp:
Understandable. Cheers! :)

PR #589, opened 2021-04-03
Improve DBus man page

exquo:
- Expand synopsis
- Add examples
- Fix typos

### DIFF
--- a/man/signal-cli-dbus.5.adoc
+++ b/man/signal-cli-dbus.5.adoc
@@ -11,10 +11,11 @@ DBus API for signal-cli - A commandline and dbus interface for the Signal messen
 
 == Synopsis
 
-*signal-cli* [--config CONFIG] daemon --dbus-system
+*signal-cli* [--verbose] [--config CONFIG] [-u USERNAME] [-o {plain-text,json}] [--dbus-system] daemon
 
-*dbus-send* --system --type=method_call --print-reply --dest="org.asamk.Signal" /org/asamk/Signal org.asamk.Signal.<method> [string:<string argument>] [array:<type>:<array argument>]
+*dbus-send* [--system | --session] [--print-reply] --type=method_call --dest="org.asamk.Signal" /org/asamk/Signal[/_<phonenum>] org.asamk.Signal.<method> [string:<string argument>] [array:<type>:<array argument>]
 
+Note: when daemon was started without explicit `-u USERNAME`, the `dbus-send` command requires adding the phone number in `/org/asamk/Signal/_<phonenum>`.
 
 == Description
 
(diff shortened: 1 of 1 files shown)


exquo:
Also, I'm not sure what `setContacts` refers to here:
https://github.com/AsamK/signal-cli/commit/caf16231aaf7550cedf74c66f5ddfa9a82b7bb88#diff-9ccadddc1e2765302bd0becc272f14ffe202fa4a08e64bf7ea3c28aabe9b3e66R129

man/signal-cli-dbus.5.adoc
```diff
@@ -11,10 +11,11 @@ DBus API for signal-cli - A commandline and dbus interface for the Signal messen
 
 == Synopsis
 
-*signal-cli* [--config CONFIG] daemon --dbus-system
+*signal-cli* [--verbose] [--config CONFIG] [-u USERNAME] [-o {plain-text,json}] [--dbus-system] daemon
```
AsamK, line 14: for system dbus it's actually `--system` after `daemon` :sweat_smile: 
`signal-cli daemon [--system]`

man/signal-cli-dbus.5.adoc
```diff
@@ -198,6 +199,19 @@ MessageReceived(timestamp<x>, sender<s>, groupId<ay>, message<s>, attachments<as
 
 This signal is received whenever we get a private message or a message is posted in a group we are an active member
 
+== Examples
+
+Send a text message (without attachment) to a contact::
+dbus-send --print-reply --type=method_call --dest="org.asamk.Signal"
```
AsamK, line 205: I think the command should be in a single line

AsamK (commented):
Thanks a lot!
I think the `setContacts` should be `setContactName`

man/signal-cli-dbus.5.adoc
exquo: I see, thanks! Should have checked the 'main' man page :). (I've never used the system bus myself)

man/signal-cli-dbus.5.adoc
exquo: Battling vim's auto-formatting :)

exquo:
> I think the setContacts should be setContactName

Makes sense. I've also added that it might be copied over from phone's contacts for a linked account.

exquo:
I've pushed another commit addressing your comments - feel free to squash with the original. 

BTW, do you have a 'recommended' way of updating PRs? Like for the PR author to do a force-push into the original branch? Or do you prefer to get separate commits (with an option to squash them later)?

AsamK:
I normally use the squash button to merge PRs, so it doesn't really matter. I don't mind force pushes to the branch.
If the PR is split into logical commits, that make the changes easier to understand I'd keep the commits, though it might be better to open separate PRs in that case.

exquo:
I see, thank you! Might be useful to know in the future.

PR #595, opened 2021-04-06
signal-cli.service adjustment for automatic startup after reboot

### LINKED ISSUE #594: The systemd unit signal-cli.service doesn't start up after reboot
The configuration file [data/signal-cli.service](/AsamK/signal-cli/blob/master/data/signal-cli.service) for systemd unit works, but the service doesn't automatically start up after reboot (at least not on Centos 7 and of course with `systemctl enable signal-cli.service` satisfied).

adaptivegarage:
The configuration file [data/signal-cli.service](data/signal-cli.service) for systemd unit works, but the service doesn't automatically start up after reboot (at least not on Centos 7 and of course with `systemctl enable signal-cli.service` satisfied).
Therefore I suggest the adjustment in this pull request, which instructs Centos 7 to start the service during standard startup scenario.
There might be further fine-tuning advisable, but I haven't had more time to go through it and this works.
Fixes #594

For anybody struggling to get this working — this is how you edit the service if you have already set it up:
edit the systemd unit conf file `signal-cli.service` on your system...
... and then reload the service configuration and recreate the symlinks that might have been created:
```
# systemctl daemon-reload
# systemctl disable signal-cli
# systemctl enable signal-cli
```

### DIFF
--- a/data/signal-cli.service
+++ b/data/signal-cli.service
@@ -13,4 +13,5 @@ User=signal-cli
 BusName=org.asamk.Signal
 
 [Install]
+WantedBy=multi-user.target
 Alias=dbus-org.asamk.Signal.service


AsamK:
This service file is meant to be dbus activatable, so in combintation with the `org.asamk.Signal.service` file it should start automatically when accessing the dbus service.

adaptivegarage:
Hi,
right, sorry, I should have read and done my research better.

cuu508:
In my testing, the service does not start on system boot, but does start the first time I try to send a message over dbus (IOW, dbus activation indeed works :-) ). 

Wouldn't there be a problem if a system reboots, but does not have a message to send for a long time, and so is also _not receiving messages for a long time_?

PR #608, opened 2021-04-30
Enable unlimited strength crypto via system Security property

i-infra:
JRE11+ supports enabling unlimited strength crypto at runtime via system property.

This factors out a possible exception pathway, ie

> WARN org.asamk.signal.util.ErrorUtils - If you use an Oracle JRE please check if you have unlimited strength crypto enabled, see README


AsamK:
Thanks!
As signal-cli now also requires java 11, I'll remove the error messages as well.

Issue #491, opened 2021-02-03
dbus mentions


Tsjippy:
Is it possible to see if a person is mentioned via dbus?

exquo:
Looks like it's a negative. I've just tried running `dbus-monitor`, sent myself a message with a mention, and it looked like this:

	signal time=1617000000.000000 sender=:1.2000 -> destination=(null destination) serial=33 path=/org/asamk/Signal; interface=org.asamk.Signal; member=MessageReceived
		 int64 1617000000000
		 string "+1234567890"
		 array of bytes [
				xx xx xx xx xx xx xx xx xx xx xx xx xx xx xx xx xx xx xx xx xx xx xx xx 
				xx xx xx xx xx xx xx xx 
		 ]
		 string "￼ message text"
		 array [
		 ]

The mention should be before the "message text", where ￼ is. (somewhat related: #409)

So, this can be a feature request. Mentions are shown in JSON and regular output though.

patcon:
From https://github.com/AsamK/signal-cli/issues/579#issuecomment-832024789, suggested change to accommodate this:

https://github.com/AsamK/signal-cli/blob/357e278f48a6e1b3e3ab58f2ae34053ab1933b96/src/main/java/org/asamk/signal/JsonDbusReceiveMessageHandler.java#L70-L75

to (with some other work):

```java
                        conn.sendMessage(new Signal.MessageReceived(objectPath,
                                message.getTimestamp(),
                                sender.getLegacyIdentifier(),
                                groupId != null ? groupId : new byte[0],
                                message.getBody().isPresent() ? message.getBody().get() : "",
                                JsonDbusReceiveMessageHandler.getAttachments(message, m), // returns json
                                JsonDbusReceiveMessageHandler.getMentions(message, m))); // NEW: returns json
```

patcon:
PR for review linked above :)

AsamK:
Mentions are now included in the new message extras of the dbus signal: https://github.com/AsamK/signal-cli/blob/master/src/main/java/org/asamk/signal/DbusReceiveMessageHandler.java#L134

PR #658, opened 2021-07-01
extend DBus signals with a field for mentions

0b11001111:
This change affects the DBus signals `SyncMessageReceived` and `MessageReceived`.
Mentions are represented as an array of structs.
Each mention holds the name of the mentioned user, its position and
length.

### DIFF
--- a/man/signal-cli-dbus.5.adoc
+++ b/man/signal-cli-dbus.5.adoc
@@ -223,7 +223,7 @@ isRegistred -> result<b>::
 
 == Signals
 
-SyncMessageReceived (timestamp<x>, sender<s>, destination<s>, groupId<ay>,message<s>, attachments<as>)::
+SyncMessageReceived (timestamp<x>, sender<s>, destination<s>, groupId<ay>, message<s>, mentions<a(sii)>, attachments<as>)::
 The sync message is received when the user sends a message from a linked device.
 
 ReceiptReceived (timestamp<x>, sender<s>)::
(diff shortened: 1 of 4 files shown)


0b11001111:
The corresponding [issue](https://github.com/AsamK/signal-cli/issues/491)

0b11001111:
What's the status of this? Also, as I'm playing around with the DBus interface and getting familiar with the project I'd like to offer improving other parts as well. However, before jumping into this I'd like to get feedback on whether this welcomed and if there any plans pointing in that direction already :)

AsamK:
Sorry for taking so long to review ... the code looks good and I like the struct :+1: 

I'd just prefer to not break existing users of the dbus interface. You can just leave the old method there as well, dbus supports method overloading.

In general the dbus interface can use some improvement, also to make it easier extensible. I'm not a dbus expert and haven't gotten around to investigate and design a better api ...

0b11001111:
I share your sentiment on breaking the API and I'll check out if I can solve this without breaking backwards compatibility. When you say the DBus API could be improved do you have anything specific in mind already?

AsamK:
For one the send command needs to be extended for new features (mentions, message replies). It would be good to have a way to extend it without having to add a new method for each new feature.
Another idea is to expose the groups and contacts as separate dbus objects.

AsamK:
I've implemented an alternative approach with an additional key/value map and new Signal names. That should extensible in the future without breaking older clients or again creating new signal names.
f39983f78acf438049cd709c17ddb3e9ccefedef

PR #666, opened 2021-07-23
add JSON output to updateGroup

technillogue:


src/main/java/org/asamk/signal/commands/UpdateGroupCommand.java
```diff
@@ -168,7 +177,13 @@ public void handleCommand(final Namespace ns, final Signal signal) throws Comman
         try {
             var newGroupId = signal.updateGroup(groupId, groupName, groupMembers, groupAvatar);
             if (groupId.length != newGroupId.length) {
-                writer.println("Created new group: \"{}\"", Base64.getEncoder().encodeToString(newGroupId));
+                String encodedGroup = Base64.getEncoder().encodeToString(newGroupId);
```
AsamK, line 180:
```suggestion
                String encodedGroupId = Base64.getEncoder().encodeToString(newGroupId);
```

src/main/java/org/asamk/signal/commands/UpdateGroupCommand.java
```diff
@@ -168,7 +177,13 @@ public void handleCommand(final Namespace ns, final Signal signal) throws Comman
...
+                if (ns.get("output") == OutputType.JSON) {
+                    final var jsonWriter = new JsonWriter(System.out);
+                    jsonWriter.write(Map.of("group", encodedGroup, "members", groupMembers, "name", groupName));
```
AsamK, line 183:
```suggestion
                    jsonWriter.write(Map.of("groupId", encodedGroupId));
```
I think outputting the groupId would be enough here, `groupMembers` and `groupName` are already provided by the user.

src/main/java/org/asamk/signal/commands/UpdateGroupCommand.java
```diff
@@ -168,7 +177,13 @@ public void handleCommand(final Namespace ns, final Signal signal) throws Comman
         try {
             var newGroupId = signal.updateGroup(groupId, groupName, groupMembers, groupAvatar);
             if (groupId.length != newGroupId.length) {
-                writer.println("Created new group: \"{}\"", Base64.getEncoder().encodeToString(newGroupId));
+                String encodedGroup = Base64.getEncoder().encodeToString(newGroupId);
+                if (ns.get("output") == OutputType.JSON) {
```
AsamK, line 181: There's a second `handleCommand` method above, please also add the json output there.

AsamK (changes requested):

src/main/java/org/asamk/signal/commands/UpdateGroupCommand.java
technillogue: if #668 is okay, I added these so that you don't need to keep state to associate group IDs with members and names.

PR #667, opened 2021-07-24
update Graal reflect config with legacy classes

technillogue:
I'm not sure if anything outdated should be removed from the config, agentlib seemed to delete most of it. Added the allowWrites as per @michaelkebe :pray:


AsamK:
Thanks!
Removing stuff shouldn't be necessary, as long as graalvm doesn't complain.

Issue #583, opened 2021-03-26
receiving and sending using same config


i0x71:
Whats the best way to send and receive messages at the same time ?
It seems you cannot do both at the same time as signal cli complains that config file is in use.

exquo:
You can run signal-cli in daemon mode 
	
	signal-cli daemon

Then add `--dbus` flag to subsequent commands, e.g.

	signal-cli --dbus send -m 'A message' +1234567890

See [man page](https://github.com/AsamK/signal-cli/blob/master/man/signal-cli.1.adoc#daemon) and [wiki](https://github.com/AsamK/signal-cli/wiki/DBus-service) for further info.

i0x71:
Thanks,
In my case i am using containers and its a bit problematic to use dbus service.
Is there any way or linking 2 signal-cli instances ?

If i generate a link with one instance and try to link a fresh one with addDevice, i get No local users found, you first need to register or link an account

i0x71:
Its rather problematic to use dbus in containers or docker, so i used linking instead.
Please close the ticket.
Thanks

maxcountryman:
I'm curious about dbus within a Docker container as well. While linking is helpful, it doesn't mitigate the config lock problem within the container itself.

waikiApp:
> I'm curious about dbus within a Docker container as well. While linking is helpful, it doesn't mitigate the config lock problem within the container itself.

What ive done is create 2 separate config folders within the container, 1 is used for sending, another is used to receiving.
Dealing with services such as systemd and dbus in container environment is a nightmare at times.

i0x71:
What ive done is create 2 separate config folders within the container, 1 is used for sending, another is used to receiving.
Dealing with services such as systemd and dbus in container environment is a nightmare at times.

maxcountryman:
> What ive done is create 2 separate config folders within the container, 1 is used for sending, another is used to receiving.
> 
> Dealing with services such as systemd and dbus in container environment is a nightmare at times.

I just did this last night and it indeed works. I think this setup probably doesn't work for everyone since your attachments etc will be in one path and not in the other config path. All things considered not too big of an issue but worth pointing out.

i0x71:
As nice as it is to have dbus, it would be ever better to have something that didnt depend on other services to interact with a daemon.

maxcountryman:
> As nice as it is to have dbus, it would be ever better to have something that didnt depend on other services to interact with a daemon.

[...]

Issue #446, opened 2021-01-17
cli interface + dbus interface in hybrid mode?


bbernhard:
Hi,

first of all, many thanks for all the hard work you are putting into this project! I've been following `signal-cli` for a while now and it's really great to see the steady improvements and new features every release. Many thanks for that - VERY much appreciated!

I am the maintainer of [signal-cli-rest-api](https://github.com/bbernhard/signal-cli-rest-api) which is a small REST API wrapper around your awesome tool. 

Right now I am spawning a new `signal-cli` process for every action. While this works, it is kinda slow due to the JAVA VM startup time (that's especially noticeable on the Raspberry Pi). I've already looked into some possible ways to improve the startup time (e.g by using [nailgun](https://github.com/facebook/nailgun)) but before I investigate further in that direction I wanted to give signal-cli's DBUS implementation a try. 

I've already played a bit with the DBUS interface and also got it working. But during my tests I stumbled across a limitation: As soon as I enable the DBUS mode, it's not possible anymore to use `signal-cli` cli interface (it blocks with `Config file is in use by another instance, waiting…`). 

I haven't looked into the implementation, but I guess it's to prevent a possible config file corruption when multiple processes are trying to write to the config? If so, would it maybe be possible to defer that check a bit and move it to the place in the code where the config file is accessed? 

What would be pretty cool is, if there would be the possibility to use the DBUS interface and the cli interface in some sort of hybrid mode (e.g configureable via config file) e.g: Only use the DBUS interface for sending messages and use the cli interface for everything else.

I know this is probably a pretty specific use case and I can totally understand if you don't want to implement such a thing (or if it's even possible to add). But before I am investigating other possible improvements (like using nailgun), I wanted to drop this idea in your repo. 

Thanks a lot for your time & have a nice day!

bentolor:
Hi @bbernhard : If it is mostly the JVM startup times. Have you already looked into using [GraalVM Native Image](https://www.graalvm.org/) to compile it to a native binary? I'd assume this would be a better addition to this project.

bbernhard:
> Hi @bbernhard : If it is mostly the JVM startup times. Have you already looked into using [GraalVM Native Image](https://www.graalvm.org/) to compile it to a native binary? I'd assume this would be a better addition to this project.

Thanks @bentolor - that looks awesome. Wasn't aware that such a thing exists!

[...]

Issue #87, opened 2017-06-15
JSON I/O event loop


kbin76:
Hi,
i've been looking into scripting signal (like making bots), but haven't found any perl or python bindings for the java library. Writing perl bindings for java seemed overwhelming for me. Also I only noted later that someone else already had started some JSON output formatting stuff, so I started off fresh on hacking together a JSON API for signal-cli. 

It's works as a form of event loop where you can post JSON requests to stdin for sending messages and it spits out JSON blobs of received messages on stdout. The API works asynchronously given the nature of the underlying message system. 

Currently only send and receive messages operations are implemented. Other operations you need to carry out using normal signal-cli commands. It's quite convenient that you can combine signal-cli command line with scripting, that way you don't need to implement all those functions in the script (like register/verify indentity that you only do once). Also it's neat to take advantage of the .config storage of signal state (trusts etc).

The fork is here:
https://github.com/kbin76/signal-cli

I've patched Main.java to implement jsonevtloop command and then created simple Perl module for interact with it as proof of concept under perl/.

Plz let me know what you think!

BR
Kalle

AsamK:
Nice work. I've just pushed my own json output format to master, so you can use that too (8717665d1d273a32afef136c15c0d5abaaae0f85)

A few suggestions:
* use the jackson json library instead of gson, no need for multiple json libraries in signal-cli
* Could be implemented as a parameter to the `daemon` command (`daemon --dbus`)

a1ad:
AsemK, 
Can you elaborate how to receive messages in json format?
Thanks

AsamK:
@a1ad I've updated the man page: https://github.com/AsamK/signal-cli/blob/master/man/signal-cli.1.adoc#receive
Just add the `--json` paramter to the receive command: `signal-cli receive --json`

kbin76:
I'll change to jackson lib instead... I started before there were any json in master and also i'm no java programmer so not sure what libs are popular etc, just grabbed one...

I'll try and take your json output and remove mine since it was a kludge anyway, but keep the json ingestion of "requests" and also try to move some stuff to have parameters enable starting as "daemon" but in json-mode as you suggested.

[...]

PR #668, opened 2021-07-24
Reading commands from standard input: now supporting every command

technillogue:
Alright, this is my latest attempt at #583 #446 #87 etc. My trick here is having the input be namespaces that would be parsed from CLI options, and then using `getCommand` the same way Main.java does it. This approach avoids the issue DBus had with feature parity. 

NamespaceDefaultingToFalse means you can avoid NullPointerException without having to set every flag. There might be a better name for this command.

### DIFF
--- a/src/main/java/org/asamk/signal/commands/Commands.java
+++ b/src/main/java/org/asamk/signal/commands/Commands.java
@@ -11,6 +11,7 @@ public class Commands {
         addCommand("addDevice", new AddDeviceCommand());
         addCommand("block", new BlockCommand());
         addCommand("daemon", new DaemonCommand());
+        addCommand("stdio", new StdioCommand());
         addCommand("getUserStatus", new GetUserStatusCommand());
         addCommand("link", new LinkCommand());
         addCommand("listContacts", new ListContactsCommand());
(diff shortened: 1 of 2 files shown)


technillogue:
@AsamK, any thoughts about this? does this go against part of signal-cli's design philosophy?

AsamK:
I like the approach.
My initial idea would be to restrict this to commands, that have json output and to maybe wrap the command output in a response object to distinguish them from incoming messages. That would probably require some refactoring for the command classes.
I will have a closer look next week.

technillogue:
How about this? Re-parsing the JSON is a little hacky, but a much smaller diff than refactoring every command. 

For which commands without JSON output would this be an issue? Although probably not super useful this does allow for keeping plain-text output by default, or specifying output format per command. 

Come to think of it, you could even do something like:

```if (inJson) {
    Map<String, Object> commandMap = jsonProcessor.readValue(input, inputType);
} else {
    var parser = App.buildArgumentParser();
    var commandMap = parser.parseArgsOrFail(input.split(" "));.
}
```

technillogue:
For which commands without JSON output would this be an issue? This does allow for keeping plain-text output by default, or specifying output format per command. 

Come to think of it, you could even do something like:

```if (inJson) {
    Map<String, Object> commandMap = jsonProcessor.readValue(input, inputType);
} else {
    var parser = App.buildArgumentParser();
    var commandMap = parser.parseArgsOrFail(input.split(" "));.
}
```
Probably not especially useful, though.

AsamK:
Thanks again for the PR.
I've now implemented this with a similar approach and an API based on JSON-RPC, see #679

Issue #671, opened 2021-07-27
DBus interface inconsistent with normal interface


johnfreed:
First, I want to add my thanks to the many who have already given them.

I have rewritten pippcat's "signalbot" program, naming mine signalmail (https://github.com/johnfreed/signalmail), to use DBus and to focus exclusively on forwarding Signal messages via email. However, the onMessageReceived function does not provide the same functionality as the main signal-cli program.

Specifically, the attachments list now consists solely of a comma-separated list of filenames. In the main program, it consists of a dictionary with more information. I know that MessageReceived *could* provide the information, because I see the following on my terminal screen when DBus is working:

  Content-Type: image/jpeg
  Type: Pointer
  Id: G8g7mvT48JAL8CSUz_ir Key length: 64
  Upload timestamp: 1627417872372 (2021-07-27T20:31:12.372Z)
  Size: 147081 bytes
  Dimensions: 1104x1472

The Id I can deduce from the filename, and the file size I suppose I could get via stat (though I would obviously like to avoid that if possible). I cannot, however, deduce the Content-Type without resorting to magic number guesswork. This is unfortunate, because that information is provided by Signal.

You may feel free to add my program to your wiki as an example of using the DBus interface.

Thanks,

John

exquo:
Cf. #316

johnfreed:
I have taken this on and would like any feedback before proceeding further. (See [https://github.com/johnfreed/signal-cli](https://github.com/johnfreed/signal-cli))

I started with the notion of retaining as much information as possible from the Signal servers, which `signal-cli` could then pass along to any end-user program like `signalmail`. I created a new data type, `DbusAttachment`, to distinguish this from the list of file names passed into `signal-cli` from the command line (I changed that data to `List<String> attachmentNames`).

Along the way I realized that the `attachmentNames` would also presumably work with arbitrary URLs, so I added in that feature and it does indeed work. So syntax like `signal-cli -u +33123456789 -a "http://my-site.com/pic.png" -m "" +440123456789` works as expected.

Conceptually, in this proposal, `signal-cli` takes attachment information from the Signal servers, packages it into a dBusAttachment, and sends that to the DBus, where it can be retrieved by the end-user program. It is important that the attachment information is guaranteed to be in a particular order, which guarantee is provided by the `Struct` class in `org.freedesktop.dbus.annotations.Position`

[...]

PR #681, opened 2021-08-11
Dbus

johnfreed:
See discussion in #671


johnfreed:
Hmm ... just saw the request for #658, which addresses a related issue, so I will take a new look at this based on the comments by @AsamK . I will separately submit the implementation of new methods for the end user.

Issue #316, opened 2020-05-30
dbus: How to get attachment filename from python


EmmanuelIT:
Using the python code for receiving messages through dbus
```python
from pydbus import SystemBus
from gi.repository import GLib

def msgRcv (timestamp, source, groupID, message, attachments):
   print("onMessageReceived:", timestamp, source, groupID, message, attachments)
   return

bus = SystemBus()
loop = GLib.MainLoop()

signal = bus.get('org.asamk.Signal')
signal.onMessageReceived = msgRcv
loop.run()
```
in **attachments** variable I get an array of strings, containing the absolute paths that the attachments reside on my filesystem. 
ex:
```['/opt/.../share/signal-cli/attachments/1309176955124639716', '/opt/.../share/signal-cli/attachments/2244858757840267913']```

How can I get the original filename of these attachments ?

The output of: 
```signal-cli --config /opt/..../share/signal-cli -u +991234567890 receive --json```
 contains a structure containing both the original (sender's) filename and the filename in my filesystem. But how can I get this structure in Python through dbus

EmmanuelIT:
Hi AsamK,

I checked  ```signal-cli/src/main/java/org/asamk/Signal.java``` and I see that the attachments are passed to the MessageReceived method as List of strings. That would normally indicate that only one "kind" of filenames can be passed: Either local filepaths where the attachments are stored or the original (the sender's) filename.
 
So is this an enhancement I am asking for, or am I doing something wrong ?

exquo:
I think this is a question to `pydbus` rather then to signal-cli, specifically on usage of the `pydbus.get(..).onMessageReceived()` function and why it has the arguments it does. Maybe you could try inspecting the dbus message objects to see what other attributes it can provide? 

Also, as you have mentioned, running
	`signal-cli -u +991234567890 daemon --json`
does have the "original" `filename` field, so you could also try parsing the json output rather then rely on pydbus.

Here's how one signal-cli client that I'm familiar with does it:
https://github.com/isamert/scli/blob/08219b56501d6139cdbdd50ba9402757ed4cfbf4/scli#L281

EmmanuelIT:
Hi exquo,

You understood correctly. 

I know I can use signal-cli command line from within python in order to get the "original" filename, but I wanted to use the dbus mechanism for obvious reasons (performance, more clear code, etc).

That's why AsamK offered the dbus mechanism in the first place.

[...]

PR #682, opened 2021-08-12
extend DBus signals for DbusAttachments 

johnfreed:
implement several subcommands for use by Dbus clients
    +    void setExpirationTimer(String number, int expiration) 
    +    boolean isRegistered(String number)
    +    List<Boolean> isRegistered(List<String> numbers)
    +    String getObjectPath()
    +    List<String> listDevices()
    +    void updateAccount()
    +    List<String> listIdentity(String number)


johnfreed:
This pull request is to address #671 #316 #658 and #491, in particular addressing the issues raised by @AsamK in #658. 

I see two possible approaches to extending the DBus interface to allow user programs to receive mentions and the full metadata surrounding an attachment. This also paves the way to implementing quotes and quoted attachments.

1) change the API. This would "bite the bullet" and basically say to any existing user that the interface was always experimental, and now it has changed. If you want your program to work, change it.

2) add to the API but maintain backward compatibility. This has the downside of some code repetition. This is the approach I take in this pull request. It maintains full backward compatibility with what I will call the legacy API while implementing a version 2. 

This PR does that by introducing two new data structures (DBus requires a `Struct` because the order of the data matters, which Java does not provide by default). The new data types are `DbusMention` (created by @0b11001111) and `DbusAttachment`.

It also introduces two new signals, `MessageReceivedV2` and `SyncMessageReceivedV2`.

The new signals provide the new data structures to the user program, while the existing signals, `MessageReceived` and `SyncMessageReceived`, continue to provide the old data types (with no information about mentions or attachment metadata).

As a result, for each message received, two signals will be emitted: the legacy signal and the V2 signal. Existing (old-style) programs will not see the V2 signal, and so nothing will break. New programs can ignore the legacy signal and process the enhanced data.

Since this is a DBus protocol, it runs on a local machine, so the environment is known. However, this approach allows extension to non-local machines, where it will not be known whether `signal-cli` is using the old or new API. The sample program, `signalmail`, demonstrates this in action, where it accepts either legacy or V2 signals and processes them appropriately. (If a V2 signal is received, then a legacy signal will always also be received, so it has logic to avoid duplicate processing of messages. If only a legacy signal is received, that means the server has an old version of signal-cli and the sample program responds accordingly.) 

The `signalmail` program also demonstrates decoding of a mention. 

As a side benefit, this PR also permits sending the contents of an arbitrary URL (not just the contents of a local file) as an attachment for both the legacy and V2 API.

johnfreed:
This pull request is to address #671 #316 #658 and #491, in particular addressing the issues raised by @AsamK in #658. 

I see two possible approaches to extending the DBus interface to allow user programs to receive mentions and the full metadata surrounding an attachment. This also paves the way to implementing quotes and quoted attachments.

1) change the API. This would "bite the bullet" and basically say to any existing user that the interface was always experimental, and now it has changed. If you want your program to work, change it.

2) add to the API but maintain backward compatibility. This has the downside of some code repetition. This is the approach I take in this pull request. It maintains full backward compatibility with what I will call the legacy API while implementing a version 2. 

This PR does that by introducing two new data structures (DBus requires a `Struct` because the order of the data matters, which Java does not provide by default). The new data types are `DbusMention` (created by @0b11001111) and `DbusAttachment`.

It also introduces two new signals, `MessageReceivedV2` and `SyncMessageReceivedV2`.

The new signals provide the new data structures to the user program, while the existing signals, `MessageReceived` and `SyncMessageReceived`, continue to provide the old data types (with no information about mentions or attachment metadata).

As a result, for each message received, two signals will be emitted: the legacy signal and the V2 signal. Existing (old-style) programs will not see the V2 signal, and so nothing will break. New programs can ignore the legacy signal and process the enhanced data.

Since this is a DBus protocol, it runs on a local machine, so the environment is known. However, this approach allows extension to non-local machines, where it will not be known whether `signal-cli` is using the old or new API. The sample program, `signalmail`, demonstrates this in action, where it accepts either legacy or V2 signals and processes them appropriately. (If a V2 signal is received, then a legacy signal will always also be received, so it has logic to avoid duplicate processing of messages. If only a legacy signal is received, that means the server has an old version of signal-cli and the sample program responds accordingly.) 

The `signalmail` program also demonstrates decoding of a mention. 

As a side benefit, this PR also permits sending the contents of an arbitrary URL (not just the contents of a local file) as an attachment for both the legacy and V2 API.

====================

Note that this pull requests also includes some older commits implementing various DBus methods. These can be cherry-picked.

0b11001111:
Fantastic! I'll check it out :)

johnfreed:
Closing this to break it into chunks as requested in CONTRIBUTING.md

PR #684, opened 2021-08-14
implement Dbus methods

johnfreed:
implement several subcommands for use by Dbus clients
+ void setExpirationTimer(String number, int expiration)
+ boolean isRegistered(String number)
+ List isRegistered(List numbers)
+ String getObjectPath()
+ List listDevices()
+ void updateAccount()
+ List listIdentity(String number)

This pull request is independent from #682 (which incorporates the same updates) and can be applied separately while #682 is under consideration.


johnfreed:
@AsamK  you should be able to cherry-pick 927065b44f5fd87415d52f1f313968dd4a89e6c1

This fixes a logic error that leaves the Dbus SignalControl methods without a SignalControl when
the daemon is called with a username. (It works fine when the daemon is called without 
a username.) The link/addDevice and register/verify subcommands rely on this.

AsamK:
That's actually on purpose. The SignalControl dbus interface is only exported when the daemon is run without a username.
The addDevice command could be implemented without a SignalCreator and the link/register/verify methods aren't that useful when running in single user mode.

johnfreed:
I'm withdrawing this pull request after reflecting on what AsamK said regarding the intentionality of having the single-user daemon not handle register, link, etc. I think philosophically that's correct, and I am redoing this work with that in mind.

I will also look at avoiding the duplication of the core routines in each of the commands, possibly along the lines of how JsonRpc was implemented.

PR #687, opened 2021-08-19
WIP: reflect config to serialize jsonrpc

technillogue:
Currently, trying to use jsonRpc on native-image produces:

```
java.lang.AssertionError: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: No serializer found for class org.asamk.signal.jsonrpc.JsonRpcResponse and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS)
```

The tracing agent doesn't notice anything about this, and my changes here don't fix the problem.

### DIFF
--- a/graalvm-config-dir/reflect-config.json
+++ b/graalvm-config-dir/reflect-config.json
@@ -487,6 +487,26 @@
   "allDeclaredConstructors":true
 },
 {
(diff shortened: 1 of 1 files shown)


AsamK:
Not sure why the agent doesn't find it, but for those classes `allDeclaredMethods` is also required so jackson can access the getter methods.

PR #703, opened 2021-08-30
add isRegistered()=false to SignalControl

bublath:
This enables the client to detect if its talking to a "control" instance or the normal instance without doing an introspective or catching exceptions.
Especially useful to distinguish a "-u" daemon from one with generic registration (where the path would be identical)


bublath:
I also added a getAccount() method (or should it rather be called getUsername?) to retrieve the current account.
This is useful to transparently support -u mode where the currently used account would not be known.

Issue #298, opened 2020-04-16
Idea: When downloading files, save them by conversation


Riamse:
Currently, attachments are downloaded to `~/.config/signal/attachments/`. In order to make things better sorted, what if we made it so that when an attachment was sent from/to +15551231234, it was saved to `~/.config/signal/attachments/+15551231234/`?

In other words: each conversation-shaped unit (e.g. a group chat or a one-on-one normal chat) should have a folder to store its corresponding attachments in.

hiqua:
signal-cli should also distinguish between accounts used, e.g. if you have two accounts send a message to the same 3rd-party account, the files shouldn't end up in the same folder.

So something like ~/.config/signal/attachments/$MY_NUMBER/$THIRD_PARTY_NUMBER would be nice.

I suspect that's trivial to implement, I'll try to do that when I find some time.

hiqua:
@AsamK what's your stance on this?

PR #707, opened 2021-09-04
Save attachments in account-specific folders

hiqua:
Related (but different): #298

I'd also like to make a PR to save in different folders for every conversation, but that looks slightly more involved so I'll open a new PR if I get around it, if that's fine for you.


johnfreed:
This seems to me like a good idea, but I'm wondering if the directory structure should be:

.local/share/signal-cli/data/PHONENUMBER.d/attachments

hiqua:
I'm open to any other naming scheme that will please @AsamK.

johnfreed:
One problem I see is backward compatibility. A message reading program (like my signalmail) should delete the attachment when the message is handled, so any such program will break.

PR #718, opened 2021-09-13
Add sendTyping and sendReceipt to dbus interface

JtheSaw:
Add DBUS implementation and documentation for both methods:
- sendTyping
- sendReceipt

### DIFF
--- a/src/main/java/org/asamk/Signal.java
+++ b/src/main/java/org/asamk/Signal.java
@@ -21,6 +21,14 @@ public interface Signal extends DBusInterface {
             String message, List<String> attachments, List<String> recipients
     ) throws Error.AttachmentInvalid, Error.Failure, Error.InvalidNumber, Error.UntrustedIdentity;
 
+    void sendTyping(
+            String recipient, Boolean stop
+    ) throws Error.Failure, Error.GroupNotFound, Error.UntrustedIdentity;
+
+    void sendReceipt(
+            String recipient, long targetSentTimestamp
+    ) throws Error.Failure, Error.UntrustedIdentity;
+
     long sendRemoteDeleteMessage(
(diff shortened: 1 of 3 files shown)


src/main/java/org/asamk/Signal.java
```diff
@@ -21,6 +21,14 @@ long sendMessage(
             String message, List<String> attachments, List<String> recipients
     ) throws Error.AttachmentInvalid, Error.Failure, Error.InvalidNumber, Error.UntrustedIdentity;
 
+    void sendTyping(
+            String recipient, Boolean stop
```
AsamK, line 25:
```suggestion
            String recipient, boolean stop
```

src/main/java/org/asamk/Signal.java
```diff
@@ -21,6 +21,14 @@ long sendMessage(
...
+    ) throws Error.Failure, Error.GroupNotFound, Error.UntrustedIdentity;
+
+    void sendReceipt(
```
AsamK, line 28: Should be renamed to `sendReadReceipt`

src/main/java/org/asamk/Signal.java
```diff
@@ -21,6 +21,14 @@ long sendMessage(
...
+            String recipient, long targetSentTimestamp
```
AsamK, line 29: Should accept a `List<Long>` of timeframe timestamps to mark multiple messages as read at the same time

AsamK (changes requested):

AsamK:
Thanks, please adapt the documentation to the review changes as well.

JtheSaw:
`List<Long>` must a new type, right?
`* <ax>  : Array of signed 64 bit integer`

AsamK (approved):

Issue #272, opened 2020-03-05
How to send group messages via dbus-send?


hasdf:
Can somebody explain how to send group messages via dbus-send?

This works great for sending to a single recipient: 

`dbus-send --session --type=method_call --print-reply --dest="org.asamk.Signal" /org/asamk/Signal org.asamk.Signal.sendMessage string:MessageText array:string: string:RECIPIENT`

I found ` void sendGroupMessage` in [Signal.java](https://github.com/AsamK/signal-cli/blob/master/src/main/java/org/asamk/Signal.java), but I'm not sure how to pass that group ID to dbus-send.

Thank you!

exquo:
Sending to a group can be done with a similar command:
```
dbus-send --session --type=method_call  --print-reply --dest=org.asamk.Signal /org/asamk/Signal org.asamk.Signal.sendGroupMessage  string:'The message goes here'  array:string:'/path/to/attachmnt1','/path/to/attachmnt2'  array:byte:139,22,72,247,116,32,170,104,205,164,207,21,248,77,185
```
where the byte array at the end is the groupId in the bytes form. It can be obtained from the list of groupIds:
```
dbus-send --session --type=method_call  --print-reply --dest='org.asamk.Signal' /org/asamk/Signal org.asamk.Signal.getGroupIds
```
To get the group name corresponding to a groupId, run
```
dbus-send --session --type=method_call  --print-reply --dest='org.asamk.Signal' /org/asamk/Signal org.asamk.Signal.getGroupName    array:byte:139,22,72,247,116,32,170,104,205,164,207,21,248,77,185     
```
Alternatively, the groupId byte array can be obtained from the base64-encoded groupId that looks like `ixZI93QgqmjNpM8V+E25==` and can be found in signal-cli's config file, by default in `~/.local/share/signal-cli/data/<PHONE_NUMBER>`. To decode it to byte array, can use python:
```
python3 -c 'import base64; print(",".join(str(i) for i in base64.b64decode(b"ixZI93QgqmjNpM8V+E25==")))'
```

exquo:
One dbus quirk here: it takes bytearrays as input (for `sendGroupMessage` and `getGroupName`) in the form `139,22,72,247,..`, but returns bytearrays (by `getGroupIds`) as space-separated hex digits `8b 16 48 f7 74 20 aa 68 cd a4 cf 15 f8 4d b9`. To convert one to the other:
```
[...]

PR #725, opened 2021-09-20
base 64 helper functions for Dbus

johnfreed:
implement alternate versions of commands that take a Group Id in
the form of byte[] to permit passing a string in base 64 format.

Methods affected:
- getGroupIds
- getGroupMembers
- getGroupName
- isGroupBlocked
- isMember
- quitGroup
- sendGroupMessage
- sendGroupMessageReaction
- sendGroupRemoteDeleteMessage
- setGroupBlocked
- updateGroup

Implement new helper function specifically for base 64 identifiers:
- getGroupIdsBase64()

Update and reorganize Dbus API documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.gitignore
 .gradle/
 .idea/*
 !.idea/codeStyles/
@@ -11,3 +12,6 @@ local.properties
 .settings/
 out/
 .DS_Store
+.asciidoctorconfig.adoc
+patches/
+signal-cli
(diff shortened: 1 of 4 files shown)


AsamK:
I'd rather not add these methods. It doubles the amount of group methods for little gain.

johnfreed:
Understood. I see from #272 a way to use external helper functions, so I'll document that on the Wiki instead.

Thanks for all your work on this. It has been interesting learning more about Signal, Dbus and Java. (Dbus on MacOSX is particularly broken, and I'll be submitting some pull requests to the Dbus maintainers, which I don't expect to see implemented, as well as Homebrew and MacPorts, where I'm more optimistic.)

PR #728, opened 2021-09-21
dbus methods: listIdentity, trust

johnfreed:
Implement new Dbus methods:
- listIdentity(number\<s\>) 
- trust(number\<s\>, safetyNumber\<s\>) 

Update documentation

### DIFF
--- a/man/signal-cli-dbus.5.adoc
+++ b/man/signal-cli-dbus.5.adoc
@@ -7,7 +7,7 @@ vim:set ts=4 sw=4 tw=82 noet:
 
 == Name
 
-DBus API for signal-cli - A commandline and dbus interface for the Signal messenger
+signal-cli-dbus - DBus API for signal-cli(1)
 
 == Synopsis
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/asamk/Signal.java
```diff
@@ -63,6 +63,10 @@ long sendGroupMessageReaction(
             String emoji, boolean remove, String targetAuthor, long targetSentTimestamp, byte[] groupId
     ) throws Error.GroupNotFound, Error.Failure, Error.InvalidNumber, Error.InvalidGroupId;
 
+    List<String> listIdentity(String number) throws Error.InvalidNumber;
```
AsamK, line 66: should rather return a struct instead of a list
And rename to `getIdentity`?

man/signal-cli-dbus.5.adoc
```diff
@@ -29,126 +29,246 @@ method(arg1<type>, arg2<type>, ...) -> return<type>
...
-* timestamp   : Can be used to identify the corresponding signal reply
+getGroupIdsBase64() -> groupList<as>::
+* groupList : Array of strings representing the internal group identifiers
 
-Depending on the type of the recipient field this sends a message to one or multiple recipients.
+Exceptions: None
 
-Exceptions: AttachmentInvalid, Failure, InvalidNumber, UntrustedIdentity
+getGroupMembers(groupId<ay>) -> members<as>::
+getGroupMembers(base64GroupId<s>) -> members<as>::
```
AsamK, line 121: Please remove unrelated doc changes here.

AsamK (changes requested):

johnfreed:
I took another approach to this one to more closely mirror the listIdentities command, and so I used that name.

Calling with an argument (a single contact) returns a dictionary just for that contact, with the contact number as key and the four data points as values.

Calling without an argument returns an array of dictionaries of all relevant contacts with the four data points. (Perhaps that should be instead a single dictionary with multiple keys?)

johnfreed:
I think I'll resubmit it as a single dictionary with multiple keys.

man/signal-cli-dbus.5.adoc
johnfreed: done

bublath:
Any chance that this gets continued? A dbus interface to restore trust is the top requirement my users would still have, as they currently have to do this from commandline, instead of using my User Interface.

AsamK:
Superseded by #1259

PR #729, opened 2021-09-21
implement Dbus isRegistered() methods

johnfreed:
isRegistered(number\<s\>) returns a boolean

isRegistered(numbers\<as\>) returns an array of Booleans

### DIFF
--- a/man/signal-cli-dbus.5.adoc
+++ b/man/signal-cli-dbus.5.adoc
@@ -7,7 +7,7 @@ vim:set ts=4 sw=4 tw=82 noet:
 
 == Name
 
-DBus API for signal-cli - A commandline and dbus interface for the Signal messenger
+signal-cli-dbus - DBus API for signal-cli(1)
 
 == Synopsis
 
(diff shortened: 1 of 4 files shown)


johnfreed:
not exactly sure why strikethrough is appearing when I use the standard java-dbus type documentation (i.e., \<s\> for string and \<as\> for array of strings)

src/main/java/org/asamk/signal/dbus/DbusSignalImpl.java
```diff
@@ -388,8 +389,28 @@ public String getGroupName(final byte[] groupId) {
     }
 
     @Override
-    public boolean isRegistered() {
-        return true;
+    public boolean isRegistered(String number) {
+        List<Boolean> result = isRegistered(List.of(number));
```
AsamK, line 393: Please use type inference, where possible
```suggestion
        var result = isRegistered(List.of(number));
```

man/signal-cli-dbus.5.adoc
```diff
@@ -29,126 +29,252 @@ method(arg1<type>, arg2<type>, ...) -> return<type>
...
+Exceptions: None
 
-sendMessage(message<s>, attachments<as>, recipient<s>) -> timestamp<x>::
-sendMessage(message<s>, attachments<as>, recipients<as>) -> timestamp<x>::
-* message     : Text to send (can be UTF8)
-* attachments : String array of filenames to send as attachments (passed as filename, so need to be readable by the user signal-cli is running under)
-* recipient   : Phone number of a single recipient
-* recipients  : Array of phone numbers 
-* timestamp   : Can be used to identify the corresponding signal reply
+getGroupIdsBase64() -> groupList<as>::
```
AsamK, line 115: please remove unrelated documentation changes

src/main/java/org/asamk/signal/dbus/DbusSignalImpl.java
```diff
@@ -388,8 +389,28 @@ public String getGroupName(final byte[] groupId) {
...
+    @Override
+    public List<Boolean> isRegistered(List<String> numbers) {
+        List<Boolean> results = new ArrayList<Boolean> ();
+        if (numbers.isEmpty()) {
+            return results;
+        }
+        try {
+            Map<String, Pair<String, UUID>> registered;
+            registered = m.areUsersRegistered(new HashSet<String>(numbers));
+            for (String number : numbers) {
```
AsamK, line 406: you can move the for loop out of the try/catch block.
And maybe convert it to a stream/map/collect construct.

src/main/java/org/asamk/Signal.java
```diff
@@ -81,7 +81,9 @@ long sendGroupMessageReaction(
             byte[] groupId, String name, List<String> members, String avatar
     ) throws Error.AttachmentInvalid, Error.Failure, Error.InvalidNumber, Error.GroupNotFound, Error.InvalidGroupId;
 
-    boolean isRegistered();
```
AsamK, line 84: please keep the `isRegistered()` method for backwards compatibility.

AsamK (changes requested):

src/main/java/org/asamk/signal/dbus/DbusSignalImpl.java
johnfreed: The stream/map/collect construct is no doubt more efficient, but I find it intimidating, like some of the Perl code I encounter from time to time. Please feel free to change what I've submitted to use it.

src/main/java/org/asamk/Signal.java
johnfreed: done

PR #759, opened 2021-10-04
Dbus get/setConfiguration methods

johnfreed:
implement:

- getConfiguration() -\> [readReceipts\<b\>, unidentifiedDeliveryIndicators\<b\>, typingIndicators\<b\>, linkPreviews\<b\>] -\> \<\>::

- setConfiguration(readReceipts\<b\>, unidentifiedDeliveryIndicators\<b\>, typingIndicators\<b\>, linkPreviews\<b\>) -\> \<\>::

Update documentation


bublath:
If this needs to get extended it will soon get very long and would require to maintain many versions of the interface to remain backwards compatible.
Wouldn't it be better to make this more flexible, e.g. name/value pairs, JSON string .... ?

AsamK:
I think for the configurations dbus properties would be a good fit.
The four properties (maybe named "ConfigurationReadReceipts") would provide read/write access to the configurations.

For the cli interface it's a single command, because new configurations can just be added as new parameters in the future.

johnfreed:
reimplemented per @AsamK's suggestion as Dbus Property list.

I did this as a new interface because only primary devices can export these properties. As a result, the interface does not exist for secondary devices.

PR #742, opened 2021-09-23
implement Dbus unregister/unlisten methods

johnfreed:
implemented two related Dbus methods:
- unregister
- unlisten

improved error handling for Dbus method:
- link

fixed logic error in Manager.java (line 876) because daemon
could still be running after a particular account is unregistered.

broadened WebSocketUnavailableException (line 913) to catch all
IOExceptions caused by unregistering account while daemon is running.

specified error message handling in ProvisioningManager.java and
DbusSignalControlImpl.java for link subcommand.

helper methods:
- getPathConfig in Manager.java
- private removeUserData in DbusSignalImpl.java

PR #761, opened 2021-10-06
Dbus listen/unlisten register/unregister

johnfreed:
This request adds new DBus functionality with `listen` and `unlisten` methods. It also handles FileLocks better so as to permit multiple instances of `signal-cli` to run concurrently handling different accounts (phone numbers) on different instances. Currently, I have tested it with two instances, one on the default system bus and a second on the session bus. 

These rely on the "anonymous" DBus daemon (as opposed to the daemon where a `-u username` is specified). In the default case, it behaves exactly as it has before: `signal-cli daemon` exports all local numbers.

But with the new `--number` option (or its synonym `--numbers`), it's possible to specify which accounts are exported on startup, and then later add and remove accounts (using `listen` and `unlisten`). In the simplest case, the daemon is started with no accounts exported: `signal-cli daemon --number`

The `unregister` method also has an optional boolean to retain the user's data; setting this to `false` erases the account information in the local data directory (normally $HOME/.local/share/signal-cli/data). 

This addresses two issues previously raised:
- "losing" messages when the daemon is started. By starting in anonymous mode with no accounts exported, the messages remain on the Signal server until the `listen` method is called. Therefore, no messages are lost.
- "locking out" accounts from the session bus when they are already exported on the system bus (or vice versa). Now the user can unlisten on one bus and then listen on the other.

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/Manager.java
+++ b/lib/src/main/java/org/asamk/signal/manager/Manager.java
@@ -94,6 +94,8 @@ public interface Manager extends Closeable {
 
     void checkAccountState() throws IOException;
 
+    SignalAccount getAccount();
+
     Map<String, Pair<String, UUID>> areUsersRegistered(Set<String> numbers) throws IOException;
 
     void updateAccountAttributes(String deviceName) throws IOException;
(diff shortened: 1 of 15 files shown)


johnfreed:
This replaces pull request #742

AsamK:
I've implemented similar functionality with `--receive-mode=manual` and `subscribeReceive`/`unsubscribeReceive`

rdtech2020:
I tried with this command then I got below error 
signal-cli: error: unrecognized arguments: '--receive-mode=manual'

PR #768, opened 2021-10-09
update docs for DBus listDevices method

johnfreed:


johnfreed:
you're too fast for me! i forgot to add the "array:" indicator for dbus-send, but it's not too important

PR #789, opened 2021-10-26
change received messages to be JsonRpcResponses instead of requests

technillogue:
It can be unexpected to receive Requests from jsonrpc server. This changes received messages to be Responses, with an id of "receive". The spec doesn't say you can't send Responses without a Request.


AsamK:
Thanks, but I think sending jsonrpc notifications is cleaner than sending unsolicited responses. Especially as the id is client defined and any id selected by the server could potentially conflict with a client id.

PR #838, opened 2021-12-16
create inet socket with wildcard address

thorsten-l:
To run signal-cli within a docker container it is much easier with a wildcard address. 

parameter `--tcp *:7583`

I just startet to create a  "docker-signal-cli" repository, https://github.com/thorsten-l/docker-signal-cli,
running signal-cli within a docker container as a daemon.


AsamK:
You can already achieve this with `--tcp [::]:12345` (or for systems without ipv6 support `--tcp 0.0.0.0:12345`)
I'd rather not add new syntax here, but it could be added to the documentation.

PR #846, opened 2021-12-23
add isRegistered to SignalControl

bublath:
To make the use of isRegistered() consistant between DbusSignal (always returns true) add a method that always returns false to DbusSignalControl.

This enables clients to easily distinguish what they're talking to


AsamK:
The `isRegistered` doesn't make sense in the SignalControl interface.
Single/multi account mode can be distinguished, by calling e.g. the `listAccounts` method and handling the `UnknownMethod` error.

AsamK:
Maybe adding a `isMultiAccount` method would make more sense?

bublath:
That would work.
However since "isRegistered" already exists, it would have been less changes to the interface and still is correct, since the SignalControl instance never is registered by itself.

PR #851, opened 2022-01-01
Added needed step for a native binary build

michaelkebe:
The libsignal_jni.so must be placed to the lib/src/main/resources/ directory to included in the native binary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,8 @@ work in all situations.
 
 1. [Install GraalVM and setup the enviroment](https://www.graalvm.org/docs/getting-started/#install-graalvm)
 2. [Install prerequisites](https://www.graalvm.org/reference-manual/native-image/#prerequisites)
-3. Execute Gradle:
+3. Copy the libsignal_jni.so to the `lib/src/main/resources/` directory 
+4. Execute Gradle:
 
         ./gradlew nativeCompile
 


AsamK:
I guess this is just necessary on systems where the provided libsignal_jni.so doesn't work, right?
On x86_64 linux this works just fine without that step.

michaelkebe:
I started with a blank docker container (ubuntu) and installed git, graalvm (with dependencies) manually. I did not installed libsignal_jni.so. And started the native binary build, which compiled successfully. But using it resulted in a library not found error. So I packed the libsignal_jni.so to the directory mentioned above.

bbernhard:
On a related note: Is anyone able to build a signal-cli native image for arm64? I recently opened an issue in the graalvm repo, because I _think_ it's  an issue with GraalVM.

michaelkebe:
I built it for x86-64. I think amd64 and x86-64 are the same.

    $ file signal-cli
    signal-cli: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=6effd96b869d7ece5fb3eac15399d7db9089b46b, for GNU/Linux 3.2.0, with debug_info, not stripped

bbernhard:
> I built it for x86-64. I think amd64 and x86-64 are the same.
> 
> ```
> $ file signal-cli
> signal-cli: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=6effd96b869d7ece5fb3eac15399d7db9089b46b, for GNU/Linux 3.2.0, with debug_info, not stripped
> ```

yeah, right. I was trying to build for arm64 - to run it on a Raspberry Pi 4. :)

AsamK:
strange, the lib should be included already ... it's configured for inclusion in the `resource-config.json` file.
https://www.graalvm.org/reference-manual/native-image/Resources/
https://github.com/AsamK/signal-cli/blob/master/graalvm-config-dir/resource-config.json#L173

@bbernhard haven't tried building for arm64 yet

bbernhard:
> strange, the lib should be included already ... it's configured for inclusion in the `resource-config.json` file. https://www.graalvm.org/reference-manual/native-image/Resources/ https://github.com/AsamK/signal-cli/blob/master/graalvm-config-dir/resource-config.json#L173
> 

For x86-64 it works for me too out of the box - without the need for copying the `libsignal_jni.so` to the `lib/src/main/resources` directory. 

(I am building the native image in a docker container too, see: https://github.com/bbernhard/signal-cli-rest-api/blob/master/Dockerfile#L65-L75)

morph027:
@bbernhard jip, working for me in https://gitlab.com/packaging/signal-cli/-/blob/master/.gitlab-ci/build.sh#L34-L76

Not as Dockerfile, but within a container using w/ `DOCKER_DEFAULT_PLATFORM=linux/arm64` ([Example build](https://gitlab.com/packaging/signal-cli/-/jobs/1933899489))

bbernhard:
> @bbernhard jip, working for me in https://gitlab.com/packaging/signal-cli/-/blob/master/.gitlab-ci/build.sh#L34-L76
> 
> Not as Dockerfile, but within a container using w/ `DOCKER_DEFAULT_PLATFORM=linux/arm64` ([Example build](https://gitlab.com/packaging/signal-cli/-/jobs/1933899489))

Thanks a lot - that sounds promising! I've just had a quick look and your solution looks pretty similar to [mine](https://github.com/bbernhard/signal-cli-rest-api/blob/c32473eeb370a6fec58f21956321169bed606b3d/Dockerfile#L61-L72) - the only dfferences I've noticed so far is, that you are modifying the `build.gradle.kts` and you are not calling `gu install native-image` :thinking:. I guess I'll have to compare it line by line and debug the issue.

morph027:
> > @bbernhard jip, working for me in https://gitlab.com/packaging/signal-cli/-/blob/master/.gitlab-ci/build.sh#L34-L76
> > Not as Dockerfile, but within a container using w/ `DOCKER_DEFAULT_PLATFORM=linux/arm64` ([Example build](https://gitlab.com/packaging/signal-cli/-/jobs/1933899489))
> 
> Thanks a lot - that sounds promising! I've just had a quick look and your solution looks pretty similar to [mine](https://github.com/bbernhard/signal-cli-rest-api/blob/c32473eeb370a6fec58f21956321169bed606b3d/Dockerfile#L61-L72) - the only dfferences I've noticed so far is, that you are modifying the `build.gradle.kts` and you are not calling `gu install native-image` thinking. I guess I'll have to compare it line by line and debug the issue.

Mine is heavily inspired by yours :) The changes to `build.gradle.kts` should not affect the native image build. I'm not even sure why `--report-unsupported-elements-at-runtime` is in there, guess it's from testing and just stayed there.

bbernhard:
> > > @bbernhard jip, working for me in https://gitlab.com/packaging/signal-cli/-/blob/master/.gitlab-ci/build.sh#L34-L76
> > > Not as Dockerfile, but within a container using w/ `DOCKER_DEFAULT_PLATFORM=linux/arm64` ([Example build](https://gitlab.com/packaging/signal-cli/-/jobs/1933899489))
> > 
> > 
> > Thanks a lot - that sounds promising! I've just had a quick look and your solution looks pretty similar to [mine](https://github.com/bbernhard/signal-cli-rest-api/blob/c32473eeb370a6fec58f21956321169bed606b3d/Dockerfile#L61-L72) - the only dfferences I've noticed so far is, that you are modifying the `build.gradle.kts` and you are not calling `gu install native-image` thinking. I guess I'll have to compare it line by line and debug the issue.
> 
> Mine is heavily inspired by yours :) The changes to `build.gradle.kts` should not affect the native image build. I'm not even sure why `--report-unsupported-elements-at-runtime` is in there, guess it's from testing and just stayed there.

I just did a few more tests - this time I used your `build.sh` script directly and ran it in a `ubunut:focal` ARM docker container on my x86-64 machine. The strange thing is, that GraalVM doesn't even start building the native image without setting env variable `JAVA_OPTS="-Djdk.lang.Process.launchMechanism=vfork"` (which you don't set in your build scripts). But even with setting the env variable, the native image build still fails :/. Next, I tried to use `docker:dind` (similar to the Gitlab CI), but unfortunately this doesn't change anything. 

But it obviously works in the Gitlab CI. So for some reason my machine and the Github CI seems to be doing something different compared to the Gitlab CI. I wonder if Gitlab is using real ARM hardware. But from what I've read online, they are probably also emulating ARM builds. Really strange.

morph027:
> But it obviously works in the Gitlab CI. So for some reason my machine and the Github CI seems to be doing something different compared to the Gitlab CI. I wonder if Gitlab is using real ARM hardware. But from what I've read online, they are probably also emulating ARM builds. Really strange.

Nope, it's not a gitlab runner, it's my private server using qemu which has been registered as runner for this project. I'll try to create a build locally.  BTW, i've run into problems using my Ryzen Desktop PC w/ qemu and graalvm, it only works on my Intel Xeon Server. Will report with the outcome later.

bbernhard:
> > But it obviously works in the Gitlab CI. So for some reason my machine and the Github CI seems to be doing something different compared to the Gitlab CI. I wonder if Gitlab is using real ARM hardware. But from what I've read online, they are probably also emulating ARM builds. Really strange.
> 
> Nope, it's not a gitlab runner, it's my private server using qemu which has been registered as runner for this project. I'll try to create a build locally. BTW, i've run into problems using my Ryzen Desktop PC w/ qemu and graalvm, it only works on my Intel Xeon Server. Will report with the outcome later.

Oh, that's interesting - many thanks for the infos! So I guess this could really be a hardware specific issue then. That could be some valuable information for the [graalvm issue](https://github.com/oracle/graal/issues/4143) :). But this probably also means that it maybe is not a GraalVM issue but a QEMU issue.

PR #852, opened 2022-01-02
Set security options for the example signal-cli-socket systemd unit

orazioedoardo:
This pull request restricts the execution environment of signal-cli (in socket mode) to make it more difficult to compromise rest of the system in case the daemon is compromised, while retaining functionality. Mainly:
1. No access to home directories
2. Read only root file system
3. Private /tmp directory
4. No access to /proc except own process
5. Socket restricted to root and users of the signal-cli group
6. Blacklist unused system calls

### DIFF
--- a/data/signal-cli-socket.service
+++ b/data/signal-cli-socket.service
@@ -5,15 +5,42 @@ After=network-online.target
 Requires=signal-cli-socket.socket
 
 [Service]
-Type=simple
+CapabilityBoundingSet=
 Environment="SIGNAL_CLI_OPTS=-Xms2m"
-ExecStart=%dir%/bin/signal-cli --config /var/lib/signal-cli daemon
-User=signal-cli
+# Update 'ReadWritePaths' if you change the config path here
+ExecStart=/usr/local/bin/signal-cli --config /var/lib/signal-cli daemon
+LockPersonality=true
+NoNewPrivileges=true
+PrivateDevices=true
+PrivateIPC=true
+PrivateTmp=true
+PrivateUsers=true
+ProcSubset=pid
+ProtectClock=true
+ProtectControlGroups=true
+ProtectHome=true
+ProtectHostname=true
+ProtectKernelLogs=true
+ProtectKernelModules=true
+ProtectKernelTunables=true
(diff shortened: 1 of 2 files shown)


data/signal-cli-socket.service
```diff
@@ -5,15 +5,43 @@ After=network-online.target
...
+RestrictSUIDSGID=true
 # JVM always exits with 143 in reaction to SIGTERM signal
 SuccessExitStatus=143
 StandardInput=socket
 StandardOutput=journal
 StandardError=journal
+SystemCallArchitectures=native
+SystemCallFilter=~@debug @mount @obsolete @privileged @resources
+UMask=0077
+# Create the user and home directory with 'useradd -r -U -s /usr/sbin/nologin -m -b /var/lib signal-cli'
```
AsamK, line 43: I've just added the systemd sysusers file (from the AUR package) to the repository, that can also be used to create the signal-cli user.

AsamK (changes requested):
Thanks!
I'd just prefer to replace the hard coded `/usr/local` path with `%dir%` again

data/signal-cli-socket.service
```diff
@@ -5,15 +5,43 @@ After=network-online.target
 Requires=signal-cli-socket.socket
 
 [Service]
-Type=simple
+CapabilityBoundingSet=
 Environment="SIGNAL_CLI_OPTS=-Xms2m"
-ExecStart=%dir%/bin/signal-cli --config /var/lib/signal-cli daemon
-User=signal-cli
+# Update 'ReadWritePaths' if you change the config path here
+ExecStart=/usr/local/bin/signal-cli --config /var/lib/signal-cli daemon
```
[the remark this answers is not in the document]
orazioedoardo: Done. Although it have to admit initially I got confused when starting the unit because `%dir%` looked like a specifier: https://man.archlinux.org/man/systemd.unit.5#SPECIFIERS

data/signal-cli-socket.service
orazioedoardo: Didn't know about that. What should I change then?

PR #879, opened 2022-01-31
Repackage signal-client native builds

### LINKED ISSUE #677: GitHub actions to build working binaries
Hello,

I've had difficulty getting 0.8.5 working on Debian 10, getting this error:
`Missing required native library dependency: libsignal-client`

I've tried compiling the dependencies, but they fail to build. I'm not a java expert, I tried googling the errors and haven't been able to figure it out.

Though I did find this repo: https://github.com/exquo/signal-libs-build
The builds there _work for me_ out of the box.

**The ask:** - What's this other guy doing differently, and can that be integrated into this repo's release builds? His work is licensed GNU GPL v3, so all of it could be used.

exquo:
# signal-cli GH Actions repackage

Fixes #677.

This workflow creates a (draft) release when a `v*` tag is pushed. It downloads the signal-client builds for windows and macos from the official repo, adds them to the signal-cli build, and uploads the resulting archives to the release assets. The "draft" status of the release means that it's only visible to the repo's maintainers, until published.

### DIFF
--- a/.github/workflows/repackage-native-libs.yml
+++ b/.github/workflows/repackage-native-libs.yml
@@ -171,13 +171,4 @@ jobs:
           if [[ "$RUNNER_OS" == 'Windows' ]]; then
             EXECUTABLE_SUFFIX=".bat"
           fi
-          if {
-            # Fail iff signal-cli complains about missing libraries.
-            # Signal-cli process's exit code is always 1, because an account is not registered yet.
-            set +o pipefail
-            ./signal-cli${EXECUTABLE_SUFFIX} receive  \
-              2>&1 |  grep 'libsignal'
-              #|&  grep 'libsignal' ;  # in bash on macos `|&` is unknown
-          }; then
-            false
-          fi
+          ./signal-cli${EXECUTABLE_SUFFIX} listAccounts
--- a/.github/workflows/repackage-native-libs.yml
+++ b/.github/workflows/repackage-native-libs.yml
@@ -1,0 +1,184 @@
+name: repackage-native-libs
+
+on:
+  push:
+    tags:
+      - v*
+
+
+jobs:
+
+  ci_wf:
+    uses: exquo/signal-cli/.github/workflows/ci.yml@master
+      # ${{ github.repository }} not accpeted here
+      # TODO: check the repo name and branch
+
+
+  lib_to_jar:
+    needs: ci_wf
+    runs-on: ubuntu-latest
+
+    outputs:
+      signal_cli_version: ${{ steps.cli_ver.outputs.signal_cli_version }}
+      release_id: ${{ steps.create_release.outputs.id }}
+
+    steps:
+
+      - name: Download signal-cli build from CI workflow
+        uses: actions/download-artifact@v2
+
+      - name: Get signal-cli version
+        id: cli_ver
+        run: |
+          #echo ${GITHUB_REF#refs/tag/}
+          tree .
+          mv ./*/*.tar.gz .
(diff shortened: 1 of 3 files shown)


.github/workflows/repackage-native-libs.yml
```diff
@@ -0,0 +1,183 @@
...
+          tree .
+
+      - name: Replace Windows lib
+        env:
+          SIGNAL_CLI_VER: ${{ steps.cli_ver.outputs.signal_cli_version }}
+          SIGNAL_CLIENT_VER: ${{ steps.lib_ver.outputs.signal_client_version }}
+        run: |
+          mv signal_jni.dll libsignal_jni.so
+          zip -u ./signal-cli-${SIGNAL_CLI_VER}/lib/signal-client-java-${SIGNAL_CLIENT_VER}.jar  ./libsignal_jni.so
+          tar -czf signal-cli-${SIGNAL_CLI_VER}-Windows.tar.gz signal-cli-${SIGNAL_CLI_VER}/
```
AsamK, line 67: Maybe it would be better to use a zip file for windows? I think there's no built-in support for tar.gz files.

.github/workflows/repackage-native-libs.yml
```diff
@@ -0,0 +1,183 @@
...
+          draft: true
+
+      - name: Upload Linux archive
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: signal-cli-${{ steps.cli_ver.outputs.signal_cli_version }}.tar.gz
+          asset_name: signal-cli-${{ steps.cli_ver.outputs.signal_cli_version }}-Linux.tar.gz
```
AsamK, line 96: I think I'd prefer the names to be lower case, then you don't need to capitalize them below.

AsamK (commented):
Nice work, thanks a lot!
I've added some suggestions, but I'd also be ok with merging it as is.

.github/workflows/repackage-native-libs.yml
```diff
@@ -0,0 +1,183 @@
...
+        run: |
+          cd signal-cli-*/bin
+          if [[ "$RUNNER_OS" == 'Windows' ]]; then
+            EXECUTABLE_SUFFIX=".bat"
+          fi
+          if {
+            # Fail iff signal-cli complains about missing libraries.
+            # Signal-cli process's exit code is always 1, because an account is not registered yet.
+            set +o pipefail
+            ./signal-cli${EXECUTABLE_SUFFIX} receive  \
```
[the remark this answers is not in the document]
exquo: Nice! That'd be a cleaner solution. I'll push the change.

.github/workflows/repackage-native-libs.yml
exquo: [Turns out](https://superuser.com/a/1431799), Win 10 now has a built-in `tar` accessible from `cmd.exe`. Also, my impression is, everyone on windows has a 7-Zip (or some similar 3rd party program) installed anyway. Not to mention, the signal-cli releases so far have been packaged as `.tar.gz`, and windows users have not been complaining :) 
But let me know, and I can change it to `.zip` if needed.

.github/workflows/repackage-native-libs.yml
AsamK: Ok, then I'd suggest to just go with the $RUNNER_OS names.

exquo: Sounds good!

AsamK:
Thanks!

PR #970, opened 2022-06-07
Native Image CI job

KevinRoebert:
Addresses https://github.com/AsamK/signal-cli/discussions/956. Windows can currently not be built due to a bug in GraalVM 22.1.0 (https://github.com/oracle/graal/issues/4502).

### DIFF
--- a/.github/workflows/native-image.yml
+++ b/.github/workflows/native-image.yml
@@ -1,0 +1,108 @@
+name: Build Native Image
+
+on:
+  push:
+    branches:
+      - '*'
+    tags:
+      - 'v*'
+
+jobs:
+  native-image:
+    env:
+      LIBSIGNAL_VERSION: 0.17.0
+    strategy:
+      fail-fast: false
+      matrix:
+        job_name:
+          - linux-amd64
+          - macos-amd64
+          #- windows-amd64
+
+        include:
+          - job_name: linux-amd64
+            os: ubuntu-latest
+
+          - job_name: macos-amd64
+            os: macOS-latest
+
+          #- job_name: windows-amd64
+          #  os: windows-latest
+      
+    name: ${{ matrix.job_name }}
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      # git checkout
+      - name: Checkout the repository
+        uses: actions/checkout@v3
+
+      # Add libs
+      - name: Add lib on *nix x86
+        run: |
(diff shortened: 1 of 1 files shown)


KevinRoebert:
Example builds can be found here: https://github.com/KevinRoebert/signal-cli/actions/runs/2457774148

.github/workflows/native-image.yml
```diff
@@ -0,0 +1,108 @@
...
+  push:
+    branches:
+      - '*'
+    tags:
+      - 'v*'
+
+jobs:
+  native-image:
+    env:
+      LIBSIGNAL_VERSION: 0.17.0
```
AsamK, line 13: would be good to have the libsignal version determined automatically, like it's done in https://github.com/AsamK/signal-cli/blob/master/.github/workflows/repackage-native-libs.yml#L40

.github/workflows/native-image.yml
```diff
@@ -0,0 +1,108 @@
...
+
+    steps:
+      # git checkout
+      - name: Checkout the repository
+        uses: actions/checkout@v3
+
+      # Add libs
+      - name: Add lib on *nix x86
+        run: |
+          curl -Lo /tmp/libsignal_jni.so.tar.gz "https://github.com/exquo/signal-libs-build/releases/download/libsignal-client_v${LIBSIGNAL_VERSION}/libsignal_jni.so-v${LIBSIGNAL_VERSION}-x86_64-unknown-linux-gnu.tar.gz"
```
AsamK, line 43: any reason why exquo's libs are used here and not the official builds, like it's done in https://github.com/AsamK/signal-cli/blob/master/.github/workflows/repackage-native-libs.yml#L51

KevinRoebert:
> Thanks, I think it would be good to have it a bit more similar to the repackage-native-libs.yml job

And what steps do you mean specifically?

For example, I tried to use the same signal-client builds (even the same job) as from the `repackage-native-libs.yml`. However, this was not properly included in the native image.
-- Reason: I do not know.

.github/workflows/native-image.yml
KevinRoebert: How to do this without downloading the packed version? Can I just read it from a file or gradle property?

Or just using always the latest version?

.github/workflows/native-image.yml
KevinRoebert: I tried to use the same signal-client builds (even the same job) as from the `repackage-native-libs.yml`. However, this was not properly included in the native image.
-- Reason: I do not know.

.github/workflows/native-image.yml
KevinRoebert: Where does the `libsignal-client-<version>.jar` file come from? Can I download it individually or find it in the repo, so I don't have to download a build every time just to get the version and then build the native image?

Issue #1022, opened 2022-10-05
DeprecatedVersionException (status code 499)


Seirdy:
Sending and receiving messages with signal-cli seems to trigger a "DeprecatedVersionException" as of 2022-10-05, 21:59 UTC. This exception seems to be intended to disable deprecated clients. I used the latest official binary release (0.11.1) and the old 0.10.0 release.

Verbose log (0.11.1):

```
2022-10-05TXX:XX:XX.XXX-XXXX [main] INFO  LibSignal - [libsignal]: rust/bridge/jni/src/logging.rs:156: Initializing libsignal version:0.20.0
2022-10-05TXX:XX:XX.XXX-XXXX [main] INFO  com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Starting...
2022-10-05TXX:XX:XX.XXX-XXXX [main] INFO  com.zaxxer.hikari.pool.HikariPool - HikariPool-1 - Added connection org.sqlite.jdbc4.JDBC4Connection@23d1e5d0
2022-10-05TXX:XX:XX.XXX-XXXX [main] INFO  com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Start completed.
2022-10-05TXX:XX:XX.XXX-XXXX [main] INFO  com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Shutdown initiated...
2022-10-05TXX:XX:XX.XXX-XXXX [main] INFO  com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Shutdown completed.
Error while checking account +XXXXXXXXXXX: StatusCode: 499
org.whispersystems.signalservice.api.push.exceptions.DeprecatedVersionException: StatusCode: 499
        at org.whispersystems.signalservice.internal.push.PushServiceSocket.validateServiceResponse(PushServiceSocket.java:1887)
        at org.whispersystems.signalservice.internal.push.PushServiceSocket.makeServiceRequest(PushServiceSocket.java:1831)
        at org.whispersystems.signalservice.internal.push.PushServiceSocket.makeServiceRequest(PushServiceSocket.java:1813)
        at org.whispersystems.signalservice.internal.push.PushServiceSocket.makeServiceBodyRequest(PushServiceSocket.java:1802)
        at org.whispersystems.signalservice.internal.push.PushServiceSocket.makeServiceRequest(PushServiceSocket.java:1738)
[...]

PR #1023, opened 2022-10-06
Update user agent

etlweather:
Recent changes from Signal caused DeprecatedVersionException to occur.

Changing the user agent debugged the receiving (sending still erroring)

Ref: 
 - https://github.com/AsamK/signal-cli/issues/1022
 - https://github.com/bbernhard/signal-cli-rest-api/issues/286
 - https://github.com/bbernhard/signal-cli-rest-api/issues/243#issuecomment-1268971240


orazioedoardo:
> (sending still erroring)

Send works for me without error too.

PR #1024, opened 2022-10-06
bump user agent version

technillogue:
stories dropped and Signal-Android/5.22.3 now seems to get DeprecatedVersionException


xrchz:
Same as #1023 ?

vaporwave9:
If this PR uses a higher version number it will also last longer, as all builds have a lifetime of 90 days:
https://github.com/signalapp/Signal-Desktop/issues/751

In theory even if there were no bugs in Signal for Android or Signal Desktop they would still have to release a new version before expiration date. @AsamK should probably merge this, if the version number works too. Else the problem will occur sooner again within the next <90 days.

technillogue:
:pray: :pray: :pray:  really appreciate handling things promptly!

PR #1078, opened 2022-10-30
Exposing Signal CLI as HTTP Server

ced-b:
Signal CLI already has a JSON RPC API, which currently uses STDIN and  STDOUT or a TCP socket. In many environments it would be best though if it was exposed as an HTTP Server. There are other projects wrapping Signal CLI, but this winds up with an unneeded dependency that someone has to maintain.

I reviewed the existing code base and saw that I could build on top of the JSON RPC functionality and expose commands as a web-server.

The below code is a very basic proof of concept which is tested for and able to send and receive single messages. It probably supports  a few other commands as well already.

The app can be fired up with:

```
signal-cli http --port 3000
```

Requests can by posting the JRPC data to:

```
http://[server]:3000/api/v1
```

It currently uses the JDK inbuild `HttpServer`, so it doesn't add the overhead of bringing in a new library.

@AsamK I'm putting this up as an initial version just to see if this would be something you would want to merge in eventually and to get any feedback technically of what would be needed. I'm still fully familiarizing myself with the code base so I may or may not implement all this in the best way possible.


AsamK:
Definitely something I'm interested in. And I like that it doesn't need an additional library.
Rather than a separate command, I would add the HTTP interface to the daemon command, as it's conceptually similar.
The daemon command currently has a `--tcp HOSTNAME:PORT` flag, so maybe `--http HOSTNAME:PORT` could start the HTTP server.
Though I guess receiving messages would be a bit difficult to handle with a simple HTTP API. Maybe a separate endpoint with Server-Sent Events (SSE) or Websockets could be used for that?

ced-b:
Ok, I could have this started from the daemon command. I will have to fully understand the code flow to make sure I'm not breaking anything. Possibly it could be forked off before everything else as it does not seem like a lot of the existing code would be needed.

As to data push, I have honestly not given that too much thought yet.

To support that I'm pretty certain I would have to bring in some lightweight server such as Undertow. An alternative that could be implemented with the inbuilt `HttpServer`, would be the good old HTTP long polling. In other words the server holds onto the HTTP request for a bit and returns the moment a message comes in. After a time out it returns anyway and the client re-initiates the connection right away.

ced-b:
I added support for additional command types and made the HTTP responses more uniform in alignment with the JSON RPC model.

Next up: add this to the Daemon command, add support for attachments and fully review what to do with push notifications.

On the attachments I noticed there is no command for this. From the documentation it seems it's expected that the attachment gets picked up from the file system. My idea is to add an attachment command that would return a JSON object with the raw bytes Base64 encoded in it.

AsamK:
Yeah, I think HTTP long polling could also work, though with the current implementation in the manager, it could have a large latency. The current receive method waits a configurable timeout for new messages, if the server has lots of new messages, all of those are received before ending the receive loop. For polling it might make sense to add some `maxMessages` parameter to limit the amount of messages received. I'll see if I can add that and to adapt the receive command for jsonrpc mode.

A command to retrieve the attachment content would make sense. I think that's independent from the HTTP interface and also useful for the other JSON-RPC interfaces, so it would be good to have a separate PR for that.

AsamK:
Yeah, I think HTTP long polling could also work, though with the current implementation in the manager, it could have a large latency. The current receive method waits a configurable timeout for new messages, if the server has lots of new messages, all of those are received before ending the receive loop. For polling it might make sense to add some `maxMessages` parameter to limit the amount of messages received. I'll see if I can add that and to adapt the receive command for jsonrpc mode.

A command to retrieve the attachment content would make sense. I think that's independent from the HTTP interface and also useful for the other JSON-RPC interfaces, so it would be good to have a separate PR for that.

AsamK:
I've extended the receive command to also work in JSON-RPC mode: 175057e781623d5e61128ccbb13f2eb47ae8bf94

The following JSON-RPC request will receive a single message, otherwise just return empty after 10 seconds.
`{"jsonrpc":"2.0","id":"id","method":"receive","params":{"account":"+XXX", "maxMessages":1,"timeout":10}}`
 In combination with the `daemon` flag `--receive-mode=manual` I think that should also work for HTTP long polling:

ced-b:
Putting a timeout on receive is a great idea. That should save me a lot of work, that I thought I would have to do.

I will create a separate merge request for the attachments shortly.

ced-b:
I move the logic for starting the HTTP server now into the `DaemonCommand`.

src/main/java/org/asamk/signal/http/HttpServerHandler.java
```diff
@@ -0,0 +1,209 @@
...
+
+    private final ObjectMapper objectMapper = Util.createJsonObjectMapper();
+
+    private final int port;
+
+    private final Manager m;
+
+    private final MultiAccountManager c;
+
+    public HttpServerHandler(final int port, final Manager m) {
```
AsamK, line 47: Just pass the `InetSocketAddress` here directly, instead of just the port, to also keep the host part.

src/main/java/org/asamk/signal/http/HttpServerHandler.java
```diff
@@ -0,0 +1,209 @@
...
+    public void init() {
+
+        try {
+
+            logger.info("Starting server on port " + port);
+
+            final var server = HttpServer.create(new InetSocketAddress(port), 0);
+            server.setExecutor(Executors.newFixedThreadPool(10));
+
+            server.createContext("/api/v1", httpExchange -> {
```
AsamK, line 68: Maybe `/api/v1/rpc` to be a bit more specific what this endpoint does.

AsamK (commented):
This should be able to reuse more of the existing JSON-RPC code. I've extracted a SignalJsonRpcCommandHandler class, that you should be able to use like this:
```java
        commandHandler = new SignalJsonRpcCommandHandler(m / c, Commands::getCommand);

        final Object[] result = {null};
        jsonRpcSender = new JsonRpcSender(s -> {
            if (result[0] != null) {
                throw new AssertionError("There should only be a single JSON-RPC response");
            }

            result[0] = s;
        });
        jsonRpcReader = new JsonRpcReader(jsonRpcSender, inputStream);
        jsonRpcReader.readMessages((method, params) -> commandHandler.handleRequest(objectMapper, method, params),
                response -> logger.debug("Received unexpected response for id {}", response.getId()));
        if (result[0] !=null) {
            sendResponse(200, result[0], httpExchange); 
        }else {
            sendResponse(201, httpExchange); 
        }
```

Other remarks:
- I think the HTTP status code should always be 200/201, unless there's a HTTP specific error or uncaught throwable.
- Would be good to set a `Content-Type: application/json` on the response and check that the request has that header as well.

ced-b:
> This should be able to reuse more of the existing JSON-RPC code. 

Funny, we both had the same idea. I just committed a change that wraps your handler. But what you proposed will be much cleaner so I will add that in.

ced-b:
Ok, your changes are now integrated.

src/main/java/org/asamk/signal/http/HttpServerHandler.java
ced-b: Makes sense. I changed that.

ced-b:
> I think the HTTP status code should always be 200/201, unless there's a HTTP specific error or uncaught throwable.

I changed this. That makes sense in the JSON RPC context as the error data is already there.

>  Would be good to set a Content-Type: application/json on the response and check that the request has that header as well.

Good point. Fixed that.

src/main/java/org/asamk/signal/http/HttpServerHandler.java
ced-b: I fixed that.

src/main/java/org/asamk/signal/http/HttpServerHandler.java
```diff
@@ -0,0 +1,123 @@
...
+        this.c = c;
+    }
+
+    public void init() {
+
+        try {
+
+            logger.info("Starting server on port " + address.toString());
+
+            final var server = HttpServer.create(new InetSocketAddress(address.getPort()), 0);
```
AsamK, line 52: You can just use the passed address here directly.
```suggestion
            final var server = HttpServer.create(address, 0);
```

src/main/java/org/asamk/signal/http/HttpServerHandler.java
```diff
@@ -0,0 +1,123 @@
...
+            server.setExecutor(Executors.newFixedThreadPool(10));
+
+            server.createContext("/api/v1/rpc", httpExchange -> {
+
+                if (!"POST".equals(httpExchange.getRequestMethod())) {
+                    sendResponse(405, null, httpExchange);
+                    return;
+                }
+
```
AsamK, line 61: Check request for Content-Type: application/json header, and respond with 415 if it doesn't match.

src/main/java/org/asamk/signal/http/HttpServerHandler.java
```diff
@@ -0,0 +1,123 @@
...
+                try {
+
+                    final SignalJsonRpcCommandHandler commandHandler;
+
+                    if (c != null) {
+                        commandHandler = new SignalJsonRpcCommandHandler(c, Commands::getCommand);
+                    } else {
+                        commandHandler = new SignalJsonRpcCommandHandler(m, Commands::getCommand);
+                    }
```
AsamK, line 70: `c` and `m` don't change, so you could actually create the CommandHandler already in the constructor.

src/main/java/org/asamk/signal/http/HttpServerHandler.java
```diff
@@ -0,0 +1,123 @@
...
+                            new JsonRpcResponse.Error(JsonRpcResponse.Error.INTERNAL_ERROR,
+                            "An internal server error has occurred.", null), null), httpExchange);
+                }
+            });
+
+            server.start();
+
+        } catch (Throwable ex) {
+            ex.printStackTrace();
+        }
```
AsamK, line 104: I don't like this outer try/catch block. If the server fails to start, signal-cli should exit with an error message.
Just make `init()` throws IOException, catch that in the DaemonCommand and rethrow it wrapped in an `IOErrorException`.

AsamK (commented):
This is shaping up nicely :)

src/main/java/org/asamk/signal/http/HttpServerHandler.java
ced-b: Oups. Fixed that.

src/main/java/org/asamk/signal/http/HttpServerHandler.java
ced-b: Makes sense. I just wasn't sure about the thread safety of ithis, so I went the safe route. But if you thing that it will be safe then that is much better and efficient.

src/main/java/org/asamk/signal/http/HttpServerHandler.java
ced-b: Yea, makes sense. Did that.

ced-b:
Ok, I fixed all the above points now.

src/main/java/org/asamk/signal/http/HttpServerHandler.java
AsamK: Good point, yes the CommandHandler is thread safe, it doesn't have any mutable state and just uses the Managers, which are thread safe.

AsamK (approved):
Thanks, looks great!

morph027:
Wow, nice work. Did you run any tests? Just grabbed latest master and tried:

```
...
signal-cli[201535]: INFO  HttpServerHandler - Starting server on localhost/127.0.0.1:3000
...

curl -vvv -H "Content-Type: application/json" -d '{"jsonrpc":"2.0", "method": "listAccounts", "id": "42"}' http://127.0.0.1:3000/api/v1
*   Trying 127.0.0.1:3000...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 3000 (#0)
> POST /api/v1 HTTP/1.1
> Host: 127.0.0.1:3000
> User-Agent: curl/7.68.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 55
> 
* upload completely sent off: 55 out of 55 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< Content-Length: 50
< Content-Type: text/html
< 
* Connection #0 to host 127.0.0.1 left intact
<h1>404 Not Found</h1>No context found for request
```

AsamK:
The endpoint in the description is not up-to-date, the path is now `/api/v1/rpc`

https://github.com/AsamK/signal-cli/blob/master/man/signal-cli.1.adoc#daemon

morph027:
Works, thank you. Looks like this will make my REST wrapper project useless, which i'm not mad about ;)

chriso0710:
Just a quick note to say thank you and well done, guys! This is a nice feature and it looks very promising!

PR #1080, opened 2022-10-31
Add command to get an attachment

ced-b:
As gone over in https://github.com/AsamK/signal-cli/pull/1078 here is a separate pull request on pulling attachment data.

The following command line command will write the Base64 string to the console:

```
signal-cli attachment --file-id=eBOlTUfsV4aD7KKXBnKr --file-name=cat.gif --content-type=image/gif
```

The following JSON RPC request will pull up the attachment in the following format:

```json
{
    "jsonrpc": "2.0",
    "method": "attachment",
    "params": {
        "fileId": "eBOlTUfsV4aD7KKXBnKr",
        "fileName": "cat.gif",
        "contentType": "image/gif"
    }
}
```

Response:

```json
{"jsonrpc":"2.0","result":{"dataBase64":"BASE64"}}
```

The actual command implementation was simple. However, the trick was to get access to the actual attachment through the attachment store. The best way I found was through the `Manager` instance passed to the command. To not have to expose all the guts of the manager etc. I wired this through the `AttachmentHelper` and to the `AttachmentStore`. Where this got a bit messy is the `DbusManagerImpl` as this would need to expose the function to get an attachment, but I did not see anything in that class for handling attachments really. Maybe you have a better idea of how to implement this.

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/AttachmentStore.java
+++ b/lib/src/main/java/org/asamk/signal/manager/AttachmentStore.java
@@ -2,8 +2,10 @@ package org.asamk.signal.manager;
 
 import org.asamk.signal.manager.util.IOUtils;
 import org.asamk.signal.manager.util.MimeUtils;
+import org.asamk.signal.manager.util.Utils;
 import org.whispersystems.signalservice.api.messages.SignalServiceAttachmentPointer;
 import org.whispersystems.signalservice.api.messages.SignalServiceAttachmentRemoteId;
+import org.whispersystems.signalservice.api.util.StreamDetails;
 
 import java.io.File;
 import java.io.FileOutputStream;
@@ -39,10 +41,12 @@ public class AttachmentStore {
                 Optional.ofNullable(pointer.getContentType()));
     }
 
-    public File getAttachmentFile(AttachmentPointer pointer) {
-        return getAttachmentFile(SignalServiceAttachmentRemoteId.from(pointer.id()),
-                Optional.ofNullable(pointer.fileName()),
-                Optional.ofNullable(pointer.contentType()));
+    public StreamDetails retrieveAttachment(final String id) throws IOException {
+        final var attachmentFile = new File(attachmentsPath, id);
+        if (!attachmentFile.exists()) {
+            return null;
+        }
+        return Utils.createStreamDetailsFromFile(attachmentFile);
     }
 
     private void storeAttachment(final File attachmentFile, final AttachmentStorer storer) throws IOException {
--- a/lib/src/main/java/org/asamk/signal/manager/AttachmentStore.java
+++ b/lib/src/main/java/org/asamk/signal/manager/AttachmentStore.java
@@ -39,6 +39,12 @@ public class AttachmentStore {
                 Optional.ofNullable(pointer.getContentType()));
     }
 
+    public File getAttachmentFile(AttachmentPointer pointer) {
+        return getAttachmentFile(SignalServiceAttachmentRemoteId.from(pointer.id()),
+                Optional.ofNullable(pointer.fileName()),
+                Optional.ofNullable(pointer.contentType()));
+    }
+
     private void storeAttachment(final File attachmentFile, final AttachmentStorer storer) throws IOException {
         createAttachmentsDir();
         try (OutputStream output = new FileOutputStream(attachmentFile)) {
--- a/src/main/java/org/asamk/signal/commands/AttachmentCommand.java
+++ b/src/main/java/org/asamk/signal/commands/AttachmentCommand.java
@@ -1,0 +1,65 @@
+package org.asamk.signal.commands;
+
+import net.sourceforge.argparse4j.inf.Namespace;
+import net.sourceforge.argparse4j.inf.Subparser;
+
+import org.asamk.signal.commands.exceptions.CommandException;
+import org.asamk.signal.commands.exceptions.UnexpectedErrorException;
+import org.asamk.signal.json.JsonAttachmentData;
+import org.asamk.signal.manager.AttachmentPointer;
+import org.asamk.signal.manager.Manager;
+import org.asamk.signal.output.JsonWriter;
+import org.asamk.signal.output.OutputWriter;
+import org.asamk.signal.output.PlainTextWriter;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Base64;
+
+public class AttachmentCommand implements JsonRpcLocalCommand {
+
+    @Override
+    public String getName() {
+        return "attachment";
+    }
+
+    @Override
+    public void attachToSubparser(final Subparser subparser) {
+        subparser.addArgument("--file-id")
+                .help("The ID of the attachment file.");
+        subparser.addArgument("--file-name")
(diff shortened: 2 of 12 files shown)


lib/src/main/java/org/asamk/signal/manager/AttachmentStore.java
```diff
@@ -39,6 +39,12 @@ public File getAttachmentFile(final SignalServiceAttachmentPointer pointer) {
                 Optional.ofNullable(pointer.getContentType()));
     }
 
+    public File getAttachmentFile(AttachmentPointer pointer) {
```
AsamK, line 42: I'd prefer to have a method here that takes the attachment id from the json output (which already includes the file extension) and returns a StreamDetails, similar to how it's done in the AvatarStore.
The Manager interface would then just return the InputStream to to the Command. And the attachment command would be responsible for closing the stream.

```java
    public StreamDetails retrieveAttachment(final String id) throws IOException {
        final var attachmentFile = new File(attachmentsPath, id);
        if (!attachmentFile.exists()) {
            return null;
        }
        return Utils.createStreamDetailsFromFile(attachmentFile);
    }

File: src/main/java/org/asamk/signal/commands/AttachmentCommand.java [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Base64;
+
+public class AttachmentCommand implements JsonRpcLocalCommand {
+
+    @Override
+    public String getName() {
+        return "attachment";
```
AsamK, line 23: That's a bit unspecific, maybe `getAttachment`?

src/main/java/org/asamk/signal/commands/AttachmentCommand.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public void attachToSubparser(final Subparser subparser) {
+        subparser.addArgument("--file-id")
+                .help("The ID of the attachment file.");
+        subparser.addArgument("--file-name")
+                .nargs("?")
+                .help("The name of the file.");
+        subparser.addArgument("--content-type")
+                .nargs("?")
+                .help("The content type of the file.");
```
AsamK, line 35: With my suggestion from above, the file-name and content-type won't be required here anymore.
Instead it would be good to add the recipient number or groupId as required parameter. Currently the attachment folder contains the attachments directly, but that might change in the future, to have them in subfolders by recipient/group.

```suggestion
        subparser.addArgument("id").required(true).help("The ID of the attachment file.");
        var mut = subparser.addMutuallyExclusiveGroup().required(true);
        mut.addArgument("--recipient").help("Sender of the attachment");
        mut.addArgument("-g", "--group-id").help("Group in which the attachment was received");
```

src/main/java/org/asamk/signal/commands/AttachmentCommand.java
```diff
@@ -0,0 +1,65 @@
...
+            final var bytes = Files.readAllBytes(file.toPath());
+            final var base64 = Base64.getEncoder().encodeToString(bytes);
+
+            if (outputWriter instanceof PlainTextWriter writer) {
+                writer.println(base64);
+            }
+            else if (outputWriter instanceof JsonWriter writer) {
+                writer.write(new JsonAttachmentData(base64));
+            }
+        } catch (IOException ex) {
```
AsamK, line 61: Would be good to catch the `FileNotFoundException` here separately and throw a UserErrorException in that case with a more specific error message.

src/main/java/org/asamk/signal/dbus/DbusManagerImpl.java
```diff
@@ -907,6 +908,12 @@ private List<MessageEnvelope.Data.Attachment> getAttachments(final Map<String, V
         }).toList();
     }
 
+    @Override
+    public File getAttachmentFile(final AttachmentPointer pointer) {
+        //TODO may need an implementation
+        return null;
```
AsamK, line 914: Just throw an exception for now, as the dbus interface doesn't support retrieving attachments yet.

```suggestion
        throw new UnsupportedOperationException();
```

src/main/java/org/asamk/signal/json/JsonAttachmentData.java
```diff
@@ -0,0 +1,3 @@
+package org.asamk.signal.json;
+
+public record JsonAttachmentData(String dataBase64) {}
```
AsamK, line 3: I don't think base64 needs to be part of the name.

```suggestion
public record JsonAttachmentData(String data) {}
```

AsamK, line 3: A future performance improvement for larger attachments might be to use an InputStream and a custom jackson serializer here to not load the complete file in memory.

AsamK (commented):
Looks good in general, I've added some comments for specifics.

Also please add the command to the man page: https://github.com/AsamK/signal-cli/blob/master/man/signal-cli.1.adoc

lib/src/main/java/org/asamk/signal/manager/AttachmentStore.java
ced-b: Fixed that.

src/main/java/org/asamk/signal/commands/AttachmentCommand.java
ced-b: Makes sense. I fixed this. I also changed the class name to match.

src/main/java/org/asamk/signal/commands/AttachmentCommand.java
ced-b: Changed that.

src/main/java/org/asamk/signal/commands/AttachmentCommand.java
ced-b: Added the catch.

src/main/java/org/asamk/signal/dbus/DbusManagerImpl.java
ced-b: I changed that

src/main/java/org/asamk/signal/json/JsonAttachmentData.java
ced-b: This was changed

ced-b: That was actually pretty straight forward to implement so I did that.

ced-b:
@AsamK I'm just writing the doc but I'm not sure that calling the parameter `recipient` makes fully sense. Would it be `originator`?

I guess generally the external number is referred to as recipient so maybe it's fine, but just wanted to throw that out.

AsamK:
> @AsamK I'm just writing the doc but I'm not sure that calling the parameter `recipient` makes fully sense. Would it be `originator`?
> 
> I guess generally the external number is referred to as recipient so maybe it's fine, but just wanted to throw that out.

Yeah, `recipient` is not the best name here, but it's also used in the other commands. So I thiink it's ok to use it here as well to keep it consistent.

AsamK (approved):
Thanks a lot, looks good!

PR #1081, opened 2022-11-02
Add a docker image

ced-b:
The attached would add a docker image for signal-cli. I had to set one up for myself so I thought it could just be added to the main project.

It can be run like this:

```
docker run AsamK/signal-cli [params]
```

The CI step pushes to DockerHub. The one for my fork can be found here but would have to be prettied up a bit: https://hub.docker.com/r/cedb2/signal-cli

You'd have to set the `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` variables.


AsamK:
Thanks!
I don't have a docker hub account. But I've taken your code as basis to publish a docker image to the github container registry:
https://github.com/AsamK/signal-cli/pkgs/container/signal-cli

ced-b:
Great. That works.

PR #1107, opened 2022-11-22
Add alive check to http daemon

ced-b:
Adds a simple HTTP endpoint that can be used by the container environment to see if the app is started and available.

The endpoint just returns HTTP status 200.


AsamK:
Thanks!

PR #1152, opened 2023-01-07
Change content-type check to check contains

ced-b:
So far it was doing an equals check, but a string like "application/json; charset=utf-8" is similarly valid. And some clients like OkHttp actually automatically add the charset.

### DIFF
--- a/src/main/java/org/asamk/signal/http/HttpServerHandler.java
+++ b/src/main/java/org/asamk/signal/http/HttpServerHandler.java
@@ -89,7 +89,7 @@ public class HttpServerHandler {
             return;
         }
 
-        if (!"application/json".equals(httpExchange.getRequestHeaders().getFirst("Content-Type"))) {
+        if (!"application/json".contains(httpExchange.getRequestHeaders().getFirst("Content-Type"))) {
             sendResponse(415, null, httpExchange);
             return;
         }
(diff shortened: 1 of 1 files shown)


etlweather:
@ced-b The condition is inverted. It won't return true in the case where content-type has the charset in it.

This would work:

```java
if (!httpExchange.getRequestHeaders().getFirst("Content-Type").contains("application/json")) {
```

ced-b:
@etlweather Good point. I fixed that now.

src/main/java/org/asamk/signal/http/HttpServerHandler.java
```diff
@@ -89,7 +89,8 @@ private void handleRpcEndpoint(HttpExchange httpExchange) throws IOException {
             return;
         }
 
-        if (!"application/json".equals(httpExchange.getRequestHeaders().getFirst("Content-Type"))) {
+        if (httpExchange.getRequestHeaders().getFirst("Content-Type") == null
+                || !httpExchange.getRequestHeaders().getFirst("Content-Type").contains("application/json")) {
```
AsamK, line 93: I think using `startsWith` instead of `contains` would be more correct.
Also please extract the `httpExchange.getRequestHeaders().getFirst("Content-Type")` to a new variable, should make this a bit more readable.

AsamK (commented):
Thanks, I agree this is overly strict at the moment.

PR #1205, opened 2023-03-20
Create signal-cli-api.go

mc2contributor:
An interface for `signal-cli daemon --socket ...` to separate command responses from incoming messages.  This also solved a problem I had with using the `--tcp` option inside a LXC container, since signal-cli binds to 127.0.0.1 and I needed access from outside the container.


AsamK:
Thanks, but I don't want that in the signal-cli repo.
Feel free to create your own repo with it and add a link to it on the [wiki](https://github.com/AsamK/signal-cli/wiki)

PR #1207, opened 2023-03-23
Added timestamp in avatar filenames to support avatar timeline

signals-from-outer-space:
Hi,

to avoid overwriting old/changed avatar files i've added a timestamp in filenames (to support avatar timeline).

Thanks a lot!


AsamK:
It's not that simple, the `retrieve` and `delete` methods don't work anymore with this change.
Also I don't see a avatar history feature in the scope of signal-cli, that could be implemented by an external script as well.

PR #1217, opened 2023-04-02
Added missing file attachment attributes in JsonAttachment output

signals-from-outer-space:
Added missing file attachment attributes in JsonAttachment output like in plaintext mode output.


AsamK:
Thanks!
I removed the `file` field`, because it's not useful in dbus socket mode. The file can be retrieved using the `getAttachment` command or by constructing the path manually.

PR #1259, opened 2023-06-04
Dbus identities

bublath:
Sorry to reopen an new PR - git still is a mystery to me sometimes and rebasing/merging code isn't working out as expected.
This PR should adress all your concerns and hints.

I'm sure there is still room for improvement.
Once we finalized this, I'll update the documentation as well of course.

### DIFF
--- a/src/main/java/org/asamk/Signal.java
+++ b/src/main/java/org/asamk/Signal.java
@@ -138,6 +138,10 @@ public interface Signal extends DBusInterface {
 
     DBusPath getDevice(long deviceId);
 
+    public DBusPath getIdentity(String number);
+    
+    public List<StructIdentity> listIdentities();
+
     List<StructDevice> listDevices() throws Error.Failure;
 
     DBusPath getThisDevice();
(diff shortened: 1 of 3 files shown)


src/main/java/org/asamk/signal/dbus/DbusSignalImpl.java
```diff
@@ -1029,6 +1033,109 @@ private void exportObject(final DBusInterface object) {
...
+
+    public class DbusSignalIdentityImpl extends DbusProperties implements Signal.Identity {
+
+        private final org.asamk.signal.manager.api.Identity identity;
+
+        public DbusSignalIdentityImpl(final org.asamk.signal.manager.api.Identity identity) {
+            this.identity=identity;
+            super.addPropertiesHandler(new DbusInterfacePropertiesHandler("org.asamk.Signal.Identity",
+                    List.of(new DbusProperty<>("number", () -> identity.recipient().getLegacyIdentifier()),
+                            new DbusProperty<>("uuid", () -> identity.recipient().getIdentifier()),
```
AsamK, line 1087:
```suggestion
                    List.of(new DbusProperty<>("number", () -> identity.recipient().number().orElse("")),
                            new DbusProperty<>("uuid", () -> identity.recipient().uuid().map(UUID::toString).orElse("")),
```

src/main/java/org/asamk/Signal.java
```diff
@@ -551,6 +555,50 @@ interface Group extends DBusInterface, Properties {
         void enableLink(boolean requiresApproval) throws Error.Failure;
     }
 
+    class StructIdentity extends Struct {
+
+        @Position(0)
+        DBusPath objectPath;
+
+        @Position(1)
+        String id;
```
AsamK, line 564: These should also be named `number` and `uuid` instead of `id`/`name`

AsamK:
The interface look good :+1:

src/main/java/org/asamk/Signal.java
bublath: Ok. Question to the interface: Should we use the uuid as Dbus Path, or the human readable number?
Personally, I would prefer the number, with fallback to the uuid when the number is not known.

bublath:
Added the documentation. While doing that I realized that case sensitivity was different from Devices and Groups, so I updated that as well.

One additional idea: it would be nice to give a similar capability as the App to display your own safety number along with a QR code, so you could establish trust both ways. Any hint how to get that and how to make a QR out of it?
(for device links I have been using https://chart.googleapis.com/chart?cht=qr&chs=200x200&chl=.<URI>)

Once we're done I probably should create one clean commit for this to clean up the mess I created when messing with git

src/main/java/org/asamk/Signal.java
AsamK: I think it's fine this way.

AsamK (approved):
Looks good, thanks!

PR #1337, opened 2023-10-05
Update signal-cli-jsonrpc.5.adoc

banderlog:
Added python example for HTTP


AsamK:
Thanks, but i don't want to add example code to the man page.
I've extended the man page with the available http endpoints.

PR #1358, opened 2023-10-28
Do not log to console when outputting as JSON

kxait:
The default logging behavior and format can cause issues when the output format is JSON, [as apparent in this issue in signal-cli-rest-api.](https://github.com/bbernhard/signal-cli-rest-api/issues/407) This PR changes the logging behavior so that if the output type is JSON, log messages won't be printed to the console (but still can be viewed using the log file)

Result:
```sh
$ ./signal-cli version 
WARN  MultiAccountManager - Ignoring +48123456789: User is not registered. (NotRegisteredException)
signal-cli 0.12.4

$ ./signal-cli --log-file log.txt -o json version
{"version":"0.12.4"}

$ cat log.txt
2023-10-28T18:28:05.124+0200 [main] WARN  o.a.s.manager.MultiAccountManager - Ignoring +48123456789: User is not registered. (NotRegisteredException)
```


AsamK:
This shouldn't be an issue, logs are printed on stderr and messages are printed on stdout.

PR #1505, opened 2024-04-13
Document the unit of "start" and "length" for mentions and text styles

brenns10:
The unit of UTF-16 code units is not necessarily obvious for users of languages that index strings by Unicode code points. Provide a pointer to an FAQ entry as well:

https://github.com/AsamK/signal-cli/wiki/FAQ#string-indexing-units

---

I admit to not compile-testing this since I don't have a development environment setup, and this was mainly documentation.


AsamK (approved):

AsamK:
Thanks a lot!

brenns10:
And thank you, signal-cli is a great project!

PR #1638, opened 2024-11-28
Rename mobile-coin-address to mobilecoin-address.

christian-oudard:
The name is MobileCoin, not "Mobile Coin".

### DIFF
--- a/src/main/java/org/asamk/signal/commands/UpdateProfileCommand.java
+++ b/src/main/java/org/asamk/signal/commands/UpdateProfileCommand.java
@@ -27,7 +27,7 @@ public class UpdateProfileCommand implements JsonRpcLocalCommand {
         subparser.addArgument("--family-name").help("New profile family name (optional)");
         subparser.addArgument("--about").help("New profile about text");
         subparser.addArgument("--about-emoji").help("New profile about emoji");
-        subparser.addArgument("--mobile-coin-address").help("New MobileCoin address (Base64 encoded public address)");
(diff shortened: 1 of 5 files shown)


AsamK:
Renaming the parameter is a backward incompatible change.
You could add it as an additional parameter, to still support the old one: `subparser.addArgument("--mobile-coin-address", "--mobilecoin-address")

christian-oudard:
I put an alias in to --mobile-coin-address. What about `cli.rs`? Would that need an alias too? I don't actually understand how the Rust code and the Java code relate to each other.

src/main/java/org/asamk/signal/commands/UpdateProfileCommand.java
```diff
@@ -27,7 +27,8 @@ public void attachToSubparser(final Subparser subparser) {
         subparser.addArgument("--family-name").help("New profile family name (optional)");
         subparser.addArgument("--about").help("New profile about text");
         subparser.addArgument("--about-emoji").help("New profile about emoji");
-        subparser.addArgument("--mobile-coin-address").help("New MobileCoin address (Base64 encoded public address)");
+        subparser.addArgument("--mobilecoin-address").help("New MobileCoin address (Base64 encoded public address)");
+        subparser.addArgument("--mobile-coin-address", "--mobilecoin-address") // Support legacy option name.
```
DonovanDiamond, line 31: You need to add a semicolon here.

aaronlumen:
I added pull #1661 which I think takes care of those items and it passes CI/CD gates as well.  Let me know if that is suitable for what you are doing or if it is totally off base from the intended results. 

Cheers - AS

PR #1674, opened 2025-01-14
Fixing RW connections deadlock on SQLite

dukhaSlayer:
Without this change we're getting a connection in the same thread we hold one already.


AsamK (approved):

AsamK:
Thanks!

Genesis-iii:
i cant even install the signal cli to ,y pc someone help me

PR #1690, opened 2025-01-22
Fixing potential handle action execution order issues.

dukhaSlayer:


dukhaSlayer (commented):
Not sure if this is a problem at the moment (hard to tell from the actions code if it's safe to execute them in random order), but might be an issue in the future.

Issue #1822, opened 2025-08-07
[Feature Request] deletelocalaccountdata for linked device


AntonKun:
I use signal-cli (in `json-rpc` mode) as a `linked device` that integrates a certain number of accounts. 
And if the user removes the integration (for example, from a mobile phone) and tries to connect to signal-cli again, an error occurs:
`"message": "The user [+************](tel:+**********) already exists\nDelete "/var/lib/signal-cli/data/******" before trying again.",`
Therefore, I have to manually clean the relevant files.

As far as I understand, the `deletelocalaccountdata `command exists precisely for deleting this data. But it seems that it is only intended for cases where signal-cli is used as the `primary device`.

Therefore, I would be very grateful if you could implement support for this command for cases where data needs to be deleted for only one account for the linked device mode.

If I am mistaken and this feature has already been implemented, please let me know the details.

Thank you in advance!

AsamK:
The `deleteLocalAccountData` command also works for linked accounts.

AntonKun:
Please advise how to use this method in json RPC mode.

I am trying to execute the following request:
POST: `/api/v1/rpc`

Body:
```json
{
    "jsonrpc": "2.0",
    "method": "deleteLocalAccountData",
    "params": {
        "account": "+380<my_number>"
    },
    "id": 1
}
```

And get:
```json
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32602,
        "message": "Method requires valid account parameter",
        "data": null
    },
    "id": 1
}
```

I also don't see in the documentation that `deleteLocalAccountData `accepts an account parameter or anything similar.

My goal is to delete local data for a specific number that was previously linked to Signal CLI using `startLink` / `finishLink`.

I use this command to run Signal CLI. Maybe there is a problem with it:
`/opt/signal-cli/bin/signal-cli daemon --http=0.0.0.0:8080 --receive-mode=manual`

PR #1851, opened 2025-09-22
Updated the JSON-RPC handler on a file lock overlap

AntonKun:
For issue #1822

**Additional description of the problem**:
The error comes from the JSON‑RPC layer trying to open a RegistrationManager for user's account, but the daemon already holds an exclusive lock on that account. This path returns "Method requires valid account parameter" when the account is "in use".

`deleteLocalAccountData `is a Registration command. Registration commands open the account file in a mode that conflicts with the daemon’s open Manager. When the lock overlaps, the server logs "Account is already in use" internally and returns that generic INVALID_PARAMS error.

**My solution**:
I updated the JSON-RPC handler so that when a registration command (like `deleteLocalAccountData`) hits a file lock, it will close the existing manager for that account and retry opening a RegistrationManager.


AsamK:
Thanks!

PR #1920, opened 2026-02-09
Add --captcha flag example to registration instructions

simonexmachina:
## Summary
- Added an inline command example showing the `--captcha` flag in the README registration section
- The existing docs mention captcha may be required and link to the wiki, but don't show the actual command syntax, which can cause users to miss this required step

## Test plan
- [ ] Verify the README renders correctly on GitHub

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,10 @@ of all country codes.)
      ```
 
   Registering may require solving a CAPTCHA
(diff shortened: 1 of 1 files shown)


AsamK:
Hm, the syntax alone is not really helpful for new users without the url to the generate the captcha.
The wiki page has both, the command syntax to register with captcha and a description how to get a captcha.

PR #1927, opened 2026-02-19
print qr code when producing a device link uri

karel1980:
Example:

image

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -88,6 +88,7 @@ dependencies {
     implementation(libs.slf4j.api)
     implementation(libs.slf4j.jul)
     implementation(libs.logback)
+    implementation(libs.zxing)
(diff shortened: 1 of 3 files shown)


AsamK:
Thanks, I like it :+1: 
I think it would be good to only print the qr code if stdout is a tty, not piped somewhere else.  To prevent breaking existing user workflows.
There doesn't seem to be a really good way in Java to do that, but checking for `System.console() != null` before printing the qr code should be good enough.

karel1980:
Updated the PR. Ciao!

AsamK (approved):

piramiday:
really cool, thanks!

related to this, would it be tough to print a similar QR code for verifying identities/contacts?
I've been going through a bunch of them in the post SPQR debacle. :wrench:

PR #1933, opened 2026-02-24
Add --no-urgent flag to send command

kaikozlov:
## Summary

- Adds a `--no-push` flag to the `send` command that sets `urgent=false` on the server request, skipping the FCM/APNs push notification to the recipient
- The message is still delivered in real-time over WebSocket if the recipient's app is active
- Follows the existing `viewOnce` pattern by adding `noPush` to the `Message` record, keeping the `Manager` interface unchanged

## Test plan

- [x] `./gradlew build` passes
- [x] `./gradlew test` passes
- [x] Manual test: send with `--no-push` and verify recipient does not receive a push notification when app is backgrounded
- [x] Manual test: send without `--no-push` and verify normal push behavior

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/api/Message.java
+++ b/lib/src/main/java/org/asamk/signal/manager/api/Message.java
@@ -13,7 +13,7 @@ public record Message(
         List<Preview> previews,
         Optional<StoryReply> storyReply,
         List<TextStyle> textStyles,
-        boolean noPush
+        boolean noUrgent
 ) {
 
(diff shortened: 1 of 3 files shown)


AsamK:
Thanks, the code looks good.
I'm not sure about the naming of the flag. The --no-push disables the push notification but the message is still `pushed` via websocket. What do you think about staying in line with the protocol naming and using something like `--no-urgent`?

kaikozlov:
Yes, I agree that `--no-urgent` is the right call on the naming. Updated!

AsamK (approved):

PR #1941, opened 2026-02-27
Fix: preserve quote context when author is null or invalid

### LINKED ISSUE #1935: Reaction field missing from SSE output in daemon --http mode
When receiving a message reaction via daemon --http SSE stream, the envelope contains:
  "dataMessage": {"timestamp": ..., "message": null, "expiresInSeconds": 0, "isExpirationUpdate": false, "viewOnce": false}

The `reaction` field is completely absent. Expected: `"dataMessage": {"reaction": {"emoji": "...", "targetSentTimestamp": ...}}`

Version: 0.13.23
Command: signal-cli -a +XX daemon --http 127.0.0.1:8080

joeykrug:
Fixes #1935

## Problem
When receiving a quoted/replied message, signal-cli silently drops the quote data if the quote author's ServiceId is null or fails the `isValid()` check. This means consumers of the signal-cli API (JSON-RPC, D-Bus) never receive quote context for these messages.

The root cause is an overly strict `.filter()` on the quote Optional in `MessageEnvelope.java` that discards the entire quote when the author can't be resolved, rather than preserving the quote with a fallback author.

## Fix
1. Remove the `.filter(q -> q.getAuthor() != null && q.getAuthor().isValid())` that drops quotes entirely
2. In `Quote.from()`, handle null/invalid authors by falling back to `RecipientAddress.UNKNOWN_UUID`
3. Use `Optional.ofNullable()` for quote text (defensive)

## Testing
This fix has been tested in production with a source-built 0.14.0-SNAPSHOT:
- Message delivery works normally (two checkmarks)
- Quote/reply context now flows through to JSON-RPC consumers
- No regressions in message sending or receiving

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/api/MessageEnvelope.java
+++ b/lib/src/main/java/org/asamk/signal/manager/api/MessageEnvelope.java
@@ -145,7 +145,6 @@ public record MessageEnvelope(
                     dataMessage.getProfileKey().isPresent(),
                     dataMessage.getReaction().map(r -> Reaction.from(r, recipientResolver, addressResolver)),
                     dataMessage.getQuote()
-                            .filter(q -> q.getAuthor() != null && q.getAuthor().isValid())
                             .map(q -> Quote.from(q, recipientResolver, addressResolver, fileProvider)),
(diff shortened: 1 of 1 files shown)


joeykrug:
Closing due to other PR being a cleaner way to do this and more comprehensive

PR #1954, opened 2026-03-02
Add reproducible builds

BarbossHack:
Ready to  merge :heavy_check_mark: 

## Purpose of this MR

Make signal-cli builds reproducible. This process lets you verify that the version of signal-cli that was downloaded from the Github Releases matches the source code in the public repository.

> Reproducible Builds ensure that the software you trust is both safe and verifiable. They do this by verifying that the binaries that you download match the original, untampered source code. For security-related tools, this means high confidence that your data and communications **are protected against hidden backdoors or vulnerabilities**.

_source: https://reproducible-builds.org/_

I have managed to make the following binaries reproducible:

- [x] JAR package (`signal-cli-XXX.tar.gz`)
- [x] Rust client binary (`signal-cli-XXX-Linux-client.tar.gz`)

The Native binary built using GraalVM is not reproducible yet and will require significantly more work. However, I think we should merge this MR first, so that people can quickly start verifying the reproducibility of signal-cli.jar in upcoming releases. I will work on making the Native binary reproducible in a follow-up MR, if that’s okay with you.

Once this has been merged, I will of course remain available to maintain and support this part of the codebase going forward.

## Description of the changes

- Add a `./reproducible-builds/` folder containing all the build and verification scripts and Dockerfiles ([similar to how Signal-Android does it](https://github.com/signalapp/Signal-Android/tree/main/reproducible-builds)).
    - `build.sh`: builds all the required artifacts: JAR, Native binary, and Rust client.
    - `verify.sh`: downloads releases from Github Releases, runs `build.sh`, and verifies whether they match the locally built ones (reproducibility check).
    - `*.Containerfile`: Dockerfiles for reproducible build, using specific tags pinned to their manifest digest hash, with pinned repository snapshot and a fixed build datetime.
    - `README.md`: contains all the reproducible build instructions for users.
- Fix the `man` Makefile to avoid adding original timestamp, which breaks reproducibility.
- Update Github Action workflows to use these new scripts:
   - In `build.yml`, use `reproducible-buidls/build.sh` to build artifacts.
   - In `release.yml`, use the artifacts from `build.yml`'s reproducible builds to create the release and containers.

## Example

For this example, I created a new tag `v0.14.0` on my fork, triggered a new workflow (see [#22589216762](https://github.com/BarbossHack/signal-cli/actions/runs/22589216762)), and verified the reproducibility like this:

```bash
./reproducible-builds/verify.sh
```


BarbossHack:
Are you interested in adding reproducible builds @AsamK? Or should I close this MR

(note that, in addition to the security benefits provided by reproducible builds, this introduces very simple scripts so that people can build signal-cli themselves the right way)

AsamK:
Thanks, I'm interested in reproducible builds. I'd just like to keep the ci build for the latest Java version as well. So the CI builds with the latest java version and the latest java LTE version

BarbossHack:
I've added a `matrix` to support building latest java version (java 26 here) :+1:

AsamK (approved):

BarbossHack:
signal-cli v0.14.3 is reproducible :partying_face: 

(I’ll keep trying to add native reproducibility, but it seems it won’t be an easy task, as GraalVM doesn’t support it natively for now)

PR #1956, opened 2026-03-03
Enforce poll choices are between 2 and 10

zmjohnso:
I noticed it was failing silently when I was trying to create a poll with more than 10 options. This PR just updates the documentation and code to reflect this hard limit of 10 poll options.


AsamK:
Thanks!

PR #1957, opened 2026-03-03
Expose a chat's isArchived status in JSON output

### LINKED ISSUE #1955: `isArchived` info in `listContacts` command is not exposed in JSON output
Hi AsamK,

`signal-cli -o json listContacts` should list the same contact info as `signal-cli listContacts --detailed` as explained by the man page:

```AsciiDoc
listContacts

Show a list of known contacts with names and profiles. When a specific recipient is given, its profile will be refreshed.

[...]

--detailed
  List the contacts with more details. If output=json, then this is always set.
```

However, this doesn't seem correct:

I noticed that info about a contact being archived is present in the `--detailed` plaintext output but it's missing with `-o json` and thus missing in JSON-RPC.

Relevant code:
https://github.com/AsamK/signal-cli/blob/master/src/main/java/org/asamk/signal/commands/ListContactsCommand.java#L93

Thanks!

moppman:
See #1955 for reference

Closes #1955

### DIFF
--- a/src/main/java/org/asamk/signal/commands/ListContactsCommand.java
+++ b/src/main/java/org/asamk/signal/commands/ListContactsCommand.java
@@ -144,6 +144,7 @@ public class ListContactsCommand implements JsonRpcLocalCommand {
                             contact.nickNameFamilyName(),
                             contact.note(),
                             contact.color(),
+                            contact.isArchived(),
                             contact.isBlocked(),
                             contact.isHidden(),
                             contact.messageExpirationTime(),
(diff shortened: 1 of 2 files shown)


moppman:
BTW, is there a way for a linked signal-cli instance to filter out contacts with no active conversation? By that I mean contacts that just kinda hang around in you contact list but you're not communicating with them via Signal, where "archived" contacts are still considered "active".

I couldn't find a way to do that. Maybe Signal does not sync this?

AsamK:
> BTW, is there a way for a linked signal-cli instance to filter out contacts with no active conversation? By that I mean contacts that just kinda hang around in you contact list but you're not communicating with them via Signal, where "archived" contacts are still considered "active".
> 
> I couldn't find a way to do that. Maybe Signal does not sync this?

No, active conversations are not synced as far as I know.

PR #1973, opened 2026-03-17
Add --voice-note flag for send command

### LINKED ISSUE #1601: [Feature Request] Add support for attachment caption
Caption is currently implemented for ReceiveMessageHandler.java:

https://github.com/AsamK/signal-cli/blob/fab1b96c21ead48b64442f6d07beedef959de6ee/src/main/java/org/asamk/signal/ReceiveMessageHandler.java#L592

And also implemented here:

https://github.com/AsamK/signal-cli/blob/fab1b96c21ead48b64442f6d07beedef959de6ee/src/main/java/org/asamk/signal/json/JsonAttachment.java#L22

While the attachment JSON format supports it, it is currently not possible to set caption for uploaded attachments as it is missing from JSON RPC, CLI options and DBUS. And the function that would need to parse the caption (if any is attached) is defined here:

https://github.com/AsamK/signal-cli/blob/master/lib/src/main/java/org/asamk/signal/manager/util/AttachmentUtils.java#L30-L40

I propose a new `--attachment-metadata` (`-am`) vector option where the user could define the additional metadata for each attachment. For example, if 4 attachments are present:

`-am [{"caption":"image1_caption"}, None, {"width":800, "height":600}, {"caption":"image4_caption", "width":800, "height":600}]`

- attachment 1 would receive caption "image1_caption"
- attachment 2 has no additional metadata
- attachment 3 has width set to 800 and height set to 600
- attachment 4 has caption "image4_caption" and has width set to 800 and height set to 600

This would allow not only to support caption but also additional metadata such as voiceNote, borderless, preview, width, height, custom_filename, etc.

KevinKickass:
## Summary

Add support for marking audio attachments as voice notes when sending messages via CLI and JSON-RPC.

Without this flag, audio attachments are sent as regular file attachments. Signal clients (especially iOS) do not provide an inline audio player for regular attachments — only voice notes get the play button UI.

## Changes

- Add `voiceNote` field to `Message` record
- Pass `voiceNote` flag through `AttachmentHelper` → `AttachmentUtils`
- Call `.withVoiceNote()` on `SignalServiceAttachmentStream` builder
- Add `--voice-note` CLI argument to `SendCommand`
- `voiceNote` parameter works automatically in JSON-RPC mode (camelCase mapping)
- DBus interface defaults to `false` (no breaking change)

## Usage

**CLI:**
```
signal-cli send -a voice.m4a --voice-note +1234567890
```

**JSON-RPC:**
```json
{"jsonrpc":"2.0","method":"send","params":{"recipient":["+1234567890"],"message":"","attachment":"voice.m4a","voiceNote":true},"id":1}
```

## Context

This addresses the TODO comment already present in `AttachmentUtils.java`:
```java
// TODO maybe add a parameter to set the voiceNote, borderless, preview, width, height and caption option
```

Closes #1601 (partially — this PR covers `voiceNote`; the other attachment metadata fields like `caption`, `borderless`, `width`, `height` could follow in a similar pattern).

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/api/Message.java
+++ b/lib/src/main/java/org/asamk/signal/manager/api/Message.java
@@ -7,6 +7,7 @@ public record Message(
         String messageText,
         List<String> attachments,
         boolean viewOnce,
+        boolean voiceNote,
         List<Mention> mentions,
         Optional<Quote> quote,
         Optional<Sticker> sticker,
--- a/lib/src/main/java/org/asamk/signal/manager/helper/AttachmentHelper.java
+++ b/lib/src/main/java/org/asamk/signal/manager/helper/AttachmentHelper.java
@@ -44,8 +44,8 @@ public class AttachmentHelper {
         return attachmentStore.retrieveAttachment(id);
     }
 
-    public List<SignalServiceAttachment> uploadAttachments(final List<String> attachments) throws AttachmentInvalidException, IOException {
-        final var attachmentStreams = createAttachmentStreams(attachments);
+    public List<SignalServiceAttachment> uploadAttachments(final List<String> attachments, boolean voiceNote) throws AttachmentInvalidException, IOException {
+        final var attachmentStreams = createAttachmentStreams(attachments, voiceNote);
 
         try {
             // Upload attachments here, so we only upload once even for multiple recipients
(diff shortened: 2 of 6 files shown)


drshields:
I came here looking exactly for this functionality.

PR #2016, opened 2026-04-13
Surface server Retry-After for rate-limit send failures

### LINKED ISSUE #1996: Idea:  429 - Rate limit
If possible - it would be beneficial to provide a timer on the 429 error, on when the rate limit will expire.  It does not  seem to be consistent, sometimes it is 1 hour, sometimes it is 12, sometimes 24.  It would be nice to know when it is safe to retry so I do not reset the clock yet again.

tonycpsu:
## Summary

- `SendMessageResult` now carries a `rateLimitRetryAfterSeconds` component populated from libsignal's `RateLimitException.getRetryAfterMilliseconds()` (for plain 413s) or `ProofRequiredException.getRetryAfterSeconds()` (for 428 challenges), giving clients a single place to read the server's advertised retry window.
- `JsonSendMessageResult.retryAfterSeconds` is now populated for every `RATE_LIMIT_FAILURE` where the server sent `Retry-After`, not only for proof-required challenges. The canonical proof-required discriminator remains `token != null`.
- Text output appends `"retry after N seconds"` to rate-limit failure lines when known.
- The aggregate `RateLimitErrorException` thrown when *all* recipients rate-limit now carries the real `nextAttemptTimestamp = now + retryAfter * 1000` instead of the hardcoded `0` placeholder.

## Motivation

Closes #1996.

Before this change, clients had no way to know when it was safe to retry a rate-limited send. Bridges, bots, and other daemons would either retry too eagerly (extending the server-side window) or wait an arbitrary amount. The data was already available upstream in libsignal — signal-cli just wasn't forwarding it.

## Design notes

- **Unified retry-after field.** Both 413 and 428 cases feed into the same `rateLimitRetryAfterSeconds` component, so downstream code doesn't need to branch on failure subtype. A new `SendMessageResults.maxRateLimitRetryAfterSeconds()` helper returns the longest wait across a batch.
- **Ceiling division for ms → s.** Plain rate-limit windows come in milliseconds; naive truncation would let clients retry up to 999ms early. Rounding up guarantees the reported window is never shorter than the server's. `SendMessageResult.millisToCeilingSeconds(long)` is a small package-private helper with its own boundary tests.
- **Source compatibility.** Adding a component widens the record's canonical constructor. A non-canonical 8-arg delegating constructor is included so pre-existing callers still compile.
- **No new dependencies.** Pure-JDK change to the existing API surface.

## Test plan

- [x] `./gradlew build` green
- [x] New `JsonSendMessageResultTest` covers: retry-after surfacing on plain 413, null retry-after when server omits Retry-After, successful sends leaving the field null, aggregate picking the longest window across a mix of populated/null entries
- [x] New `SendMessageResultTest` covers: ms→s ceiling boundaries (0, 1, 500, 999, 1000, 1001, 1500, 60000), and the 8-arg legacy constructor still compiles and produces a null retry-after
- [x] Live send smoke-tested end-to-end (successful send returns `type: SUCCESS` with no retryAfterSeconds, as expected)

The JSON contract extension (populating `retryAfterSeconds` for plain 413s where it was previously null) is documented in `CHANGELOG.md` under Unreleased.

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/api/SendMessageResult.java
+++ b/lib/src/main/java/org/asamk/signal/manager/api/SendMessageResult.java
@@ -15,6 +15,32 @@ public record SendMessageResult(
         Long rateLimitRetryAfterSeconds
 ) {
 
+    /**
+     * Source-compatible constructor for callers built against the pre-retry-after record shape.
+     * Delegates to the canonical constructor with a null retry-after, which is the correct value
+     * for any result not produced by {@link #from}.
+     */
+    public SendMessageResult(
+            RecipientAddress address,
+            boolean isSuccess,
+            boolean isNetworkFailure,
+            boolean isUnregisteredFailure,
+            boolean isIdentityFailure,
+            boolean isRateLimitFailure,
+            ProofRequiredException proofRequiredFailure,
+            boolean isInvalidPreKeyFailure
+    ) {
+        this(address,
+                isSuccess,
+                isNetworkFailure,
+                isUnregisteredFailure,
+                isIdentityFailure,
+                isRateLimitFailure,
+                proofRequiredFailure,
+                isInvalidPreKeyFailure,
+                null);
+    }
+
     public static SendMessageResult unregisteredFailure(RecipientAddress address) {
         return new SendMessageResult(address, false, false, true, false, false, null, false, null);
     }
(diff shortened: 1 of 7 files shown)


lib/src/main/java/org/asamk/signal/manager/api/SendMessageResult.java
```diff
@@ -11,28 +11,71 @@ public record SendMessageResult(
...
-                sendMessageResult.isInvalidPreKeyFailure());
+                rateLimitFailure != null || proofRequiredFailure != null,
+                proofRequiredFailure == null ? null : new ProofRequiredException(proofRequiredFailure),
+                sendMessageResult.isInvalidPreKeyFailure(),
+                retryAfterSeconds);
+    }
+
+    static long millisToCeilingSeconds(long millis) {
+        // Round up so we never advise a retry before the server's deadline.
+        return (millis + 999L) / 1000L;
```
AsamK: ceil

AsamK:
Thanks!

PR #2019, opened 2026-04-14
Fix sender key re-distribution on every group message

### LINKED ISSUE #2018: signal-cli daemon: sender key re-distributed on every group message, causing 6s delay per send
When running `signal-cli daemon --dbus` in multi-account mode, every `sendGroupMessage` DBus call opens a fresh unidentified (sealed sender) connection and re-sends the sender key distribution message to all group members - even when all recipients already have the current distribution key. This causes a ~6 second delay on every group message.

## Environment

- **signal-cli version:** 0.14.2
- **Mode:** `daemon --dbus`, multi-account
- **Platform:** Debian testing-slim (Docker/Podman container, linux/amd64)
- **Java:** `default-jre-headless` from Debian testing (openjdk 25)

## Steps to reproduce

1. Start signal-cli in daemon mode: `signal-cli --config /var/lib/signal-cli daemon --dbus`
2. Call `sendGroupMessage` via DBus
3. Note the ~6 second delay and the log output
4. Immediately call `sendGroupMessage` again on the same group
5. Observe the same ~6 second delay again

## Expected behaviour

After the sender key distribution has been sent to all group members, subsequent messages to the same group should reuse the existing distribution without re-sending it, and without re-opening the unidentified connection.

## Actual behaviour

Every `sendGroupMessage` call:
1. Logs `Need to send the distribution message to N addresses` - even for recipients who already have the current distribution
2. Opens a fresh unidentified (sealed sender) connection (`[LibSignalChatConnection]: [unidentified:...] Connecting...`)
3. Waits ~6 seconds for the unidentified connection to establish
4. Sends the distribution message
5. Sends the actual group message
6. Updates the `sender_key_shared` table with a new timestamp for the same `distribution_id`

## Log evidence

Two consecutive `sendGroupMessage` calls, each taking ~6.5 seconds:

```
[DBus-MethodCall-Receiver-2] WARN  o.a.s.manager.internal.ManagerImpl - No profile name set. ...
[DBus-MethodCall-Receiver-2] DEBUG o.a.signal.manager.helper.SendHelper - Can use sender key for 2/2 recipients.
[DBus-MethodCall-Receiver-2] INFO  LibSignal - [SignalServiceMessageSender]: [sendGroupMessage][1776174600586] Need to send the distribution message to 2 addresses.
[RxCachedThreadScheduler-1]  INFO  LibSignalService - [LibSignalChatConnection]: [unidentified:198703640] Connecting...
[RxCachedThreadScheduler-1]  INFO  LibSignalService - [LibSignalChatConnection]: [sendRequest] Enqueuing request send for after connection
[RxCachedThreadScheduler-5]  INFO  LibSignalService - [LibSignalChatConnection]: [sendRequest] Enqueuing request send for after connection
[Thread-29]                  INFO  LibSignalService - [LibSignalChatConnection]: [unidentified:198703640] Connected
[DBus-MethodCall-Receiver-2] INFO  LibSignal - [SignalServiceMessageSender]: [sendGroupMessage][1776174600586] Successfully sent sender keys to 2/2 recipients.
[DBus-MethodCall-Receiver-2] DEBUG o.a.signal.manager.helper.SendHelper - Successfully sent using sender key to 2/2 sender key targets.
[DBus-MethodCall-Receiver-2] DEBUG o.a.signal.manager.helper.SendHelper - Sending took PT6.581S
```

The same sequence repeats for the immediately following message.

## Database analysis

Inspection of the SQLite account database (`account.db`) confirms:

The sender key is not being rotated. The same `distribution_id` is used across all sends:

```sql
SELECT hex(distribution_id), datetime(created_timestamp/1000,'unixepoch')
FROM sender_key
WHERE address = '44da2512-df79-4fec-a3e6-a6ffb1cdc5a9';
-- EBCFEEAB4AE5473FB1380D3F6FBADCDE | 2026-04-14 13:49:06
```

The `sender_key_shared` table is updated on every send, with the same `distribution_id` but a new timestamp:

| Send | distribution_id | timestamp |
|------|----------------|-----------|
| 1st  | `EBCFEEAB4AE5...` | 2026-04-14 14:08:45 |
| 2nd  | `EBCFEEAB4AE5...` | 2026-04-14 15:02:45 |

All 5 recipient device entries (2 recipients, 5 devices total) are updated with the new timestamp on every call - confirming that signal-cli re-distributes to all devices every time, despite them already having the current key.

**Schema:**

```sql
CREATE TABLE sender_key_shared (
  _id INTEGER PRIMARY KEY,
  address TEXT NOT NULL,
  device_id INTEGER NOT NULL,
  distribution_id BLOB NOT NULL,
  timestamp INTEGER NOT NULL,
  UNIQUE(address, device_id, distribution_id)
) STRICT;
```

The `UNIQUE(address, device_id, distribution_id)` constraint means only one entry per recipient/device/distributionId combination exists - signal-cli is doing an upsert (updating the timestamp) rather than skipping the distribution when the record already exists.

## Root cause hypothesis

It appears that signal-cli's check for "does this recipient need the sender key distribution?" may not be correctly skipping recipients whose `sender_key_shared` record already exists with the current `distribution_id`. Instead it appears to always re-distribute, then upsert the record with a new timestamp.

Additionally, the unidentified (sealed sender) connection is not being pooled or kept alive between DBus calls in daemon mode. Each `sendGroupMessage` call opens a fresh TLS connection to Signal's unidentified delivery service, which takes ~6 seconds. Signal's mobile clients maintain a persistent unidentified connection - the same approach would eliminate this delay in daemon mode.

## Impact

Every group message sent via `signal-cli daemon --dbus` takes 6+ seconds due to:
1. Re-distribution of already-known sender keys
2. A fresh unidentified connection setup for each send

smeinecke:
Fixes #2018

### Problem

When sending group messages via any signal-cli interface (DBus daemon, JSON-RPC, CLI),
every `sendGroupMessage` call re-distributes the sender key to all group members — even
when they already have the current distribution. This causes a ~6 second delay per message
because each re-distribution opens a fresh unidentified (sealed sender) TLS connection.

Log output observed on every send, including immediate consecutive messages:

```
DEBUG Can use sender key for 2/2 recipients.
INFO  Need to send the distribution message to 2 addresses.   ← every time
INFO  [LibSignalChatConnection]: [unidentified:...] Connecting...
# ~6 seconds later
INFO  Successfully sent sender keys to 2/2 recipients.
DEBUG Sending took PT6.581S
```

### Root cause

`SenderKeySharedStore.markSenderKeysSharedWith()` stores addresses using
`entry.toString()` instead of `entry.address()`.

Since `SenderKeySharedEntry` is a Java record, `toString()` returns the full record
representation rather than just the UUID:

```
# Stored in DB (wrong):
SenderKeySharedEntry[address=e32b08b1-0b1a-4869-a264-a0b7b8bb68ad, deviceId=1]

# Expected:
e32b08b1-0b1a-4869-a264-a0b7b8bb68ad
```

When signal-service-java subsequently calls `getSenderKeySharedWith()` and compares the
retrieved addresses against the actual group member UUIDs, the comparison always fails.
signal-service-java therefore believes the distribution has never been sent and re-sends
it on every `sendGroupMessage` call.

This can be confirmed by inspecting the SQLite `sender_key_shared` table — the `address`
column contains the full record `toString()` representation:

```sql
SELECT address FROM sender_key_shared LIMIT 1;
-- SenderKeySharedEntry[address=e32b08b1-0b1a-4869-a264-a0b7b8bb68ad, deviceId=1]
```

### Fix

One-character change in `SenderKeySharedStore.markSenderKeysSharedWith()`:

```java
// Before (wrong — stores full record toString())
statement.setString(1, entry.toString());

// After (correct — stores just the UUID address)
statement.setString(1, entry.address());
```

### Affected send modes

All send modes route through the same `SenderKeySharedStore` and are equally affected:

```
CLI send command
JSON-RPC (--socket / --http)
DBus daemon
       ↓
SignalServiceMessageSender.sendGroupDataMessage()
       ↓
SenderKeySharedStore.markSenderKeysSharedWith()  ← fix applied here
```

### Verified

After applying the fix, consecutive group messages no longer trigger
"Need to send the distribution message":

```
# Message 1 (first send after restart — distribution needed once):
DEBUG Can use sender key for 2/2 recipients.
INFO  Need to send the distribution message to 2 addresses.
DEBUG Sending took PT7.164S

# Message 2 (immediately after):
DEBUG Can use sender key for 2/2 recipients.
DEBUG Successfully sent using sender key to 2/2 sender key targets.
DEBUG Sending took PT6.059S
```

### Known remaining issue

A ~6s delay still occurs on every message due to a separate problem: the unidentified
(sealed sender) connection is torn down a few seconds after each send and must be
re-established for the next message:

```
[unidentified:1636243480] Connected      ← message sent
[unidentified:1636243480] Disconnecting  ← closed ~4s later
[unidentified:255738755]  Connecting     ← next message, new 6s wait
```

The authenticated connection is kept alive via periodic keep-alives; the unidentified
connection is not. Signal mobile clients maintain a persistent unidentified connection.
This is a separate issue in libsignal's connection management and is outside the scope
of this fix.

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/helper/SendHelper.java
+++ b/lib/src/main/java/org/asamk/signal/manager/helper/SendHelper.java
@@ -705,6 +705,17 @@ public class SendHelper {
                     successCount,
                     addresses.size());
 
+            final var successfulAddresses = results.stream()
+                    .filter(SendMessageResult::isSuccess)
+                    .flatMap(r -> r.getSuccess()
+                            .getDevices()
+                            .stream()
+                            .map(device -> new SignalProtocolAddress(r.getAddress().getIdentifier(), device)))
+                    .toList();
+            if (!successfulAddresses.isEmpty()) {
+                account.getSenderKeyStore().markSenderKeySharedWith(distributionId, successfulAddresses);
+            }
+
             return results;
         } catch (org.whispersystems.signalservice.api.crypto.UntrustedIdentityException e) {
             return null;
--- a/lib/src/main/java/org/asamk/signal/manager/storage/senderKeys/SenderKeySharedStore.java
+++ b/lib/src/main/java/org/asamk/signal/manager/storage/senderKeys/SenderKeySharedStore.java
@@ -204,7 +204,7 @@ public class SenderKeySharedStore {
         ).formatted(TABLE_SENDER_KEY_SHARED);
         try (final var statement = connection.prepareStatement(sql)) {
             for (final var entry : newEntries) {
-                statement.setString(1, entry.toString());
+                statement.setString(1, entry.address());
                 statement.setInt(2, entry.deviceId());
                 statement.setBytes(3, UuidUtil.toByteArray(distributionId.asUuid()));
                 statement.setLong(4, System.currentTimeMillis());


AsamK (approved):
Thanks a lot!

PR #2021, opened 2026-04-15
Add distinct JSON-RPC error code for captcha rejection

tonycpsu:
## Problem

`submitRateLimitChallenge` currently maps `CaptchaRejectedException` to the generic `USER_ERROR` code (`-1`) in the JSON-RPC layer. This makes it impossible for callers to programmatically distinguish a rejected captcha token from any other user error (missing params, unknown method, etc.) without fragile string-matching on the error message.

## Solution

Introduce `CaptchaRejectedErrorException` and a dedicated error code (`-6` / `CAPTCHA_REJECTED_ERROR`). The change follows the same pattern as the existing `RateLimitErrorException` (`-5`).

**Before:**
```json
{ "error": { "code": -1, "message": "Captcha rejected, it may be outdated..." } }
```

**After:**
```json
{ "error": { "code": -6, "message": "Captcha rejected, it may be outdated..." } }
```

The error message is unchanged. The CLI exit code for this path becomes `6`, consistent with the per-error-type exit code convention.

## Changes

- `CaptchaRejectedErrorException.java` — new exception type mirroring `RateLimitErrorException`
- `CommandException.java` — adds `CaptchaRejectedErrorException` to the sealed `permits` clause
- `SubmitRateLimitChallengeCommand.java` — throws `CaptchaRejectedErrorException` instead of `UserErrorException`
- `SignalJsonRpcCommandHandler.java` — maps it to code `-6`
- `Main.java` — handles it in the CLI exit code switch (exit `6`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/man/signal-cli.1.adoc
+++ b/man/signal-cli.1.adoc
@@ -1168,6 +1168,7 @@ signal-cli -a ACCOUNT trust -a RECIPIENT
 * *3*: Server or IO error
 * *4*: Sending failed due to untrusted key
 * *5*: Server rate limiting error
+* *6*: CAPTCHA was rejected
 
 == Files
 
--- a/src/main/java/org/asamk/signal/Main.java
+++ b/src/main/java/org/asamk/signal/Main.java
@@ -22,6 +22,7 @@ import net.sourceforge.argparse4j.impl.Arguments;
 import net.sourceforge.argparse4j.inf.ArgumentParserException;
 import net.sourceforge.argparse4j.inf.Namespace;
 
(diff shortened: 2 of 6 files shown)


AsamK:
Please update the documentation as well https://github.com/AsamK/signal-cli/blob/master/man/signal-cli.1.adoc#exit-codes

tonycpsu:
> Please update the documentation as well https://github.com/AsamK/signal-cli/blob/master/man/signal-cli.1.adoc#exit-codes

Good catch, will do.

AsamK (approved):

PR #2034, opened 2026-04-28
fix: flush SSE response headers immediately on connect

### LINKED ISSUE #2033: SSE response on /api/v1/events is delayed until the first keepalive.
SSE response on /api/v1/events is delayed until the first keepalive.

  - signal-cli 0.14.3
  - JDK 25
  - normal java

## Steps to Reproduce
1. Start signal-cli in HTTP daemon mode.
2. Run:
   `curl -N -sS -H 'Accept: text/event-stream' -D - -o /dev/null -w '\nstarttransfer=%{time_starttransfer}s\n' 'http://127.0.0.1:7583/api/v1/events'`
3. Observe that the initial response is delayed until the first keepalive.

## Expected Behavior
The SSE endpoint should flush the initial HTTP 200 response immediately after the connection is established.

For the `curl` command above, the expected result is:

| Case | Status | First SSE chunk | `starttransfer` |
|---|---|---|---|
| with fix | `200` | `:\n` | `~14 ms` |

## Actual Behavior
The initial response is delayed until the first keepalive is sent about 15 seconds later, so clients with shorter header timeouts can fail before receiving any bytes.

For the `curl` command above, the observed result on the unfixed version is:

| Case | Status | First SSE chunk | `starttransfer` |
|---|---|---|---|
| without fix | `200` | `:\n` | `~15017 ms` |

legacycode:
## Summary

This fixes a delay in the SSE endpoint exposed by `daemon --http`.
`/api/v1/events` previously did not flush the initial HTTP response until a later write in the 15 second keepalive loop. Clients with shorter header timeouts could fail before receiving the initial HTTP 200 response.
Closes #2033.

## Root cause
In `HttpServerHandler.handleEventsEndpoint()`, the code sends the HTTP response headers and creates the `ServerSentEventSender`, but does not flush the response before entering the 15 second keepalive loop.
Because of that, the JVM HTTP server buffers the response until a later write flushes it.

## Fix
Flush the response body immediately after creating the `ServerSentEventSender`.
This preserves the existing keepalive behavior and only ensures that the initial HTTP response is sent to the client immediately.

## Test
Adds an integration test covering the SSE endpoint behavior.

The test verifies that:
- the endpoint responds with `200`
- the initial HTTP response is returned within `2` seconds
- the response is not delayed until the first keepalive interval

### DIFF
--- a/src/main/java/org/asamk/signal/http/HttpServerHandler.java
+++ b/src/main/java/org/asamk/signal/http/HttpServerHandler.java
@@ -169,6 +169,12 @@ public class HttpServerHandler implements AutoCloseable {
             httpExchange.sendResponseHeaders(200, 0);
             final var sender = new ServerSentEventSender(httpExchange.getResponseBody());
 
+            // Flush HTTP response headers to the client immediately.
+            // Without this, the JVM HttpServer buffers everything until a later write
+            // in the keep-alive loop (15 s), causing clients with shorter timeouts
+            // (e.g. 10 s) to abort before receiving the initial response.
+            httpExchange.getResponseBody().flush();
+
             final var shouldStop = new AtomicBoolean(false);
             final var handlers = subscribeReceiveHandlers(managers, sender, () -> {
                 shouldStop.set(true);
--- a/src/test/java/org/asamk/signal/http/SseInitialFlushTest.java
+++ b/src/test/java/org/asamk/signal/http/SseInitialFlushTest.java
@@ -1,0 +1,537 @@
+package org.asamk.signal.http;
+
+import org.asamk.signal.manager.Manager;
+import org.asamk.signal.manager.api.CallInfo;
+import org.asamk.signal.manager.api.CallOffer;
+import org.asamk.signal.manager.api.Configuration;
+import org.asamk.signal.manager.api.Device;
+import org.asamk.signal.manager.api.DeviceLinkUrl;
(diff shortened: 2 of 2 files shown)


src/main/java/org/asamk/signal/http/HttpServerHandler.java
```diff
@@ -169,6 +169,12 @@ private void handleEventsEndpoint(HttpExchange httpExchange) throws IOException
             httpExchange.sendResponseHeaders(200, 0);
             final var sender = new ServerSentEventSender(httpExchange.getResponseBody());
 
+            // Flush HTTP response headers to the client immediately.
+            // Without this, the JVM HttpServer buffers everything until the first
+            // write+flush in the keep-alive loop (15 s), causing clients with shorter
+            // timeouts (e.g. 10 s) to abort before receiving any data.
+            sender.sendKeepAlive();
```
AsamK: Hm, would just calling `httpExchange.getResponseBody().flush()` work as well, to send out headers?

legacycode: I tested that locally, and `httpExchange.getResponseBody().flush()` is sufficient to fix the actual timeout issue here.

I updated the PR accordingly and adjusted the test to verify that the initial HTTP response is returned within 2 seconds, instead of requiring an immediate SSE keepalive byte.

AsamK (approved):

PR #2068, opened 2026-06-11
Fix NPE in endCall when deviceId is null

neagix:
When there are issues with the tunnel, it might be better to fail more gracefully.


AsamK:
Thanks, even if no offer was exchanged yet, a hangup message should be sent.
I checked Signal-Android and that just sends 0 in that field so I've changed it to do that here as well.

PR #2069, opened 2026-06-11
Clean up PulseAudio virtual devices after tunnel exits

neagix:
I noticed that the virtual devices are accumulating and not getting cleaned up; verified with this patch that cleanup works.

See commit message for more details.

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/helper/CallManager.java
+++ b/lib/src/main/java/org/asamk/signal/manager/helper/CallManager.java
@@ -409,6 +409,9 @@ public class CallManager implements AutoCloseable {
             // Monitor process exit
             process.onExit().thenAcceptAsync(p -> {
                 logger.debug("Tunnel for call {} exited with code {}", callIdUnsigned(state.callId), p.exitValue());
+                // SIGTERM (code 143) kills the Rust process without running Drop,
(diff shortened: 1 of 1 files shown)


lib/src/main/java/org/asamk/signal/manager/helper/CallManager.java
```diff
@@ -409,6 +409,9 @@ private void spawnMediaTunnel(CallState state) {
             // Monitor process exit
             process.onExit().thenAcceptAsync(p -> {
                 logger.debug("Tunnel for call {} exited with code {}", callIdUnsigned(state.callId), p.exitValue());
+                // SIGTERM (code 143) kills the Rust process without running Drop,
```
AsamK, line 412: SIGTERM gives the Rust process the chance to clean up, so this should be fixed in [signal-call-tunnel](https://github.com/visigoth/signal-call-tunnel) to handle SIGTERM signal correctly.

neagix: I will try opening a PR there then.

neagix: [Here](https://github.com/visigoth/signal-call-tunnel/pull/3).

PR #2077, opened 2026-06-30
Skip known-unregistered recipients when sending to groups

KritBlade:
## Why

When sending to a group, signal-cli builds the recipient list from the **full** group membership and attempts delivery to every member. Members that are already known to be unregistered (they left Signal / deleted their account) are never excluded, so they are retried on **every** send.

For a group that has accumulated many such dead accounts, this forces signal-cli off the fast sealed-sender group path onto the legacy per-recipient (1:1) fan-out, and each unregistered recipient costs a failed network round-trip.

## What it breaks

On a real 143-member group containing **35 unregistered** members, a single send — even a typing indicator — took **~23.7 s**:

```
Successfully sent using 1:1 to 2/41 legacy targets.
Sending took PT23.704S
```

This is bad on its own, but it becomes a hard failure under `daemon --http`: the single-threaded daemon is blocked for the full ~23 s, a client/health-check timeout then tears the daemon down mid-send, the message is never delivered, and the daemon restarts — repeating for every send to that group. (There is no JVM crash / `hs_err`; it's a slow-send-gets-killed loop, which made it hard to diagnose.)

## The fix

Before sending a group message, filter out recipients whose `unregistered_timestamp` is already set, returning an `unregisteredFailure` result for each **without** the network attempt. Result reporting is unchanged — those recipients still appear as `UNREGISTERED_FAILURE` in the send results and CLI output; they are simply no longer dialed.

This is safe because the unregistered flag is maintained **independently** by profile/CDS discovery (`ProfileHelper`), which clears it via `markRegistered(..., true)` when a recipient registers again — so skipped recipients are automatically re-included. The send path itself neither sets nor clears the flag, so nothing is permanently blacklisted.

`sendGroupMessageInternal` is the single choke point that both data messages and typing indicators funnel through, so one filter covers every group send.

### Changes
- **`RecipientStore#getUnregisteredRecipientIds(Set<RecipientId>)`** — new query returning the subset of the given recipients that have `unregistered_timestamp` set.
- **`SendHelper#sendGroupMessageInternal`** — partition out known-unregistered recipients up front and emit their `unregisteredFailure` results without a network send.

## Testing

Built with `./gradlew installDist` (JDK 25) and exercised against the same 143-member group:

| | before | after |
|---|---|---|
| internal send time | `PT23.704S` | `PT2.405S` |
| `Hit unregistered user` network attempts | 35 | **0** |
| daemon RPC `sendTyping` | `HTTP 000`, daemon killed mid-send | `HTTP 200` in ~1.6 s, daemon survives |
| group send results | — | 104 `SUCCESS` / 35 `UNREGISTERED_FAILURE` / 3 `IDENTITY_FAILURE` |

Verbose output confirms the new path: `Skipping 35 known-unregistered recipient(s) in group send.`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+
+- Sending to large groups is no longer slowed down by members that are already known to be unregistered; they are skipped instead of being retried via the legacy 1:1 send path on every send.
+
 ## [0.14.5] - 2026-06-11
 
 ### Changed
--- a/lib/src/main/java/org/asamk/signal/manager/helper/SendHelper.java
+++ b/lib/src/main/java/org/asamk/signal/manager/helper/SendHelper.java
@@ -507,9 +507,32 @@ public class SendHelper {
     ) throws IOException {
         long startTime = System.currentTimeMillis();
 
-        final var addressesMap = recipientIds.stream()
+        // Recipients that are already known to be unregistered are skipped here.
+        // Otherwise every group send re-attempts them via the slow legacy 1:1
+        // fan-out, which on large groups can take tens of seconds (and time out).
+        // The unregistered flag is maintained independently by profile/CDS
+        // discovery, which clears it once a recipient registers again, so they
+        // are re-included automatically. An unregisteredFailure result is still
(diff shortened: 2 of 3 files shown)


AsamK (approved):
Thanks!

PR #2079, opened 2026-07-03
Fix HTTP handler to accept ACI/UUID account parameter in SSE endpoint

### LINKED ISSUE #2078: HTTP 400 on SSE /api/v1/events when ?account= is passed with ACI/UUID instead of phone number
## Bug Description

When running `signal-cli` in multi-account daemon mode (without `-u` flag) and calling the SSE events endpoint `/api/v1/events` with `?account=<ACI-UUID>`, the server returns HTTP 400 with an empty body.

This happens because `MultiAccountManagerImpl.getManager(String)` only looks up accounts by phone number (`m.getSelfNumber()`). When an ACI/UUID (e.g. `cc528f93-527e-4566-8c62-d12dc99dbce0`) is passed instead of a phone number, the lookup fails and `getManagerFromQuery()` returns null, which causes the handler to respond with HTTP 400.

## Steps to Reproduce

1. Link a device to signal-cli in multi-account mode
2. Start the daemon: `signal-cli daemon --http :8080`
3. Call the SSE endpoint: `curl -N "http://localhost:8080/api/v1/events?account=<aci-uuid>"`
4. Server returns HTTP 400

## Environment

- signal-cli: 0.14.5 (normal Java)
- OS: Linux (Debian 12, kernel 6.1)
- JDK: OpenJDK 21

tilllt:
## Description

`MultiAccountManagerImpl.getManager()` only looked up accounts by phone number, causing HTTP 400 errors when the SSE events endpoint was called with `?account=<ACI-UUID>` (e.g., `cc528f93-527e-4566-8c62-d12dc99dbce0`).

This is a common scenario when automation tools identify accounts by their Signal ACI/UUID rather than phone number.

### `SignalAccountFiles.java`
- Added `initManagerByAci(String aciStr)` method that looks up accounts via `AccountsStore.getPathByAci()` instead of by phone number

### `MultiAccountManagerImpl.java`
- `getManager()` now first tries phone number lookup, then falls back to ACI/UUID lookup when the phone number lookup fails

### `HttpServerHandler.java`
- `getManagerFromQuery()` now logs a warning when the account identifier is not found and falls back to returning all managers instead of returning null (which caused HTTP 400)

## Related Issue

Closes #2078

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/internal/MultiAccountManagerImpl.java
+++ b/lib/src/main/java/org/asamk/signal/manager/internal/MultiAccountManagerImpl.java
@@ -95,23 +95,42 @@ public class MultiAccountManagerImpl implements MultiAccountManager {
     }
 
     @Override
-    public Manager getManager(final String number) {
+    public Manager getManager(final String identifier) {
         synchronized (managers) {
-            final var manager = managers.stream()
-                    .filter(m -> m.getSelfNumber().equals(number))
+            // Try to find already loaded manager by phone number
+            var manager = managers.stream()
+                    .filter(m -> m.getSelfNumber().equals(identifier))
                     .findFirst()
                     .orElse(null);
             if (manager != null) {
                 return manager;
             }
+
+            // Try to load by phone number first
             try {
-                final var newManager = signalAccountFiles.initManager(number);
+                final var newManager = signalAccountFiles.initManager(identifier);
                 managers.add(newManager);
                 return newManager;
-            } catch (IOException | NotRegisteredException | AccountCheckException e) {
-                logger.warn("Failed to load new manager", e);
+            } catch (NotRegisteredException e) {
+                // Not a valid phone number or not registered yet, try ACI
+                logger.debug("Manager not found by number, trying ACI: {}", identifier);
+            } catch (IOException | AccountCheckException e) {
+                logger.warn("Failed to load new manager by number: {}", identifier, e);
                 return null;
             }
+
+            // Try to load by ACI (useful for SSE endpoint with ?account=<UUID>)
+            try {
+                final var newManager = signalAccountFiles.initManagerByAci(identifier);
+                managers.add(newManager);
+                return newManager;
+            } catch (NotRegisteredException e) {
+                logger.debug("Manager not found by ACI: {}", identifier);
+            } catch (IOException | AccountCheckException e) {
+                logger.warn("Failed to load new manager by ACI: {}", identifier, e);
+            }
+
+            return null;
         }
     }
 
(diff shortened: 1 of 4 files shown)


tilllt:
## Update: Additional fixes for the SSE endpoint

After testing the initial fix, I discovered and resolved three more issues:

### 1. `Util.getQueryMap()` — `+` decoded to space by `URLDecoder`
`URLDecoder.decode()` converts `+` to space per URL spec, but phone numbers legitimately contain `+`. Now escapes `+` to `%2B` before decoding, so `?account=+4915...` works correctly.

### 2. `IllegalArgumentException` not caught in `getManager()`
When `initManager()` is called with a UUID (not a phone number), it can throw `IllegalArgumentException`. Added to the catch blocks alongside `IOException | AccountCheckException`.

### 3. `OverlappingFileLockException` for UUID lookups on already-loaded accounts
When the daemon starts, `initMultiAccountManager()` loads all accounts. But a subsequent `getManager(UUID)` call only checked loaded managers by phone number — not by UUID. So it fell through to `initManagerByAci()` → `initManager()` → `SignalAccount.load()`, which tried to lock a file already locked by the startup load.

**Fix**: Added `getAccountNumberByAci()` helper that looks up UUID in `accountsStore.getAllAccounts()`. Before attempting `initManagerByAci()`, `getManager()` now checks if the UUID corresponds to an already-loaded manager (by resolving UUID → phone number → existing manager in the loaded list).

### Verification
All SSE endpoint variants tested successfully (`%2B`, UUID, empty, bare `+`, plain number — all return HTTP 200).

src/main/java/org/asamk/signal/http/HttpServerHandler.java
```diff
@@ -262,7 +262,10 @@ private List<Manager> getManagerFromQuery(final Map<String, String> query) {
             } else {
                 final var manager = c.getManager(account);
                 if (manager == null) {
-                    return null;
+                    // Account not found by the given identifier (number or ACI/UUID)
+                    // Log the available accounts to help debug
+                    logger.warn("Account not found for identifier: {}", account);
+                    return c.getManagers();
```
AsamK: if a specific account was requested but not found, then it shouldn't return all managers.
Should stay `return null` here.

lib/src/main/java/org/asamk/signal/manager/internal/MultiAccountManagerImpl.java
```diff
@@ -95,23 +95,60 @@ public void addOnManagerRemovedHandler(final Consumer<Manager> handler) {
     }
 
     @Override
-    public Manager getManager(final String number) {
+    public Manager getManager(final String identifier) {
         synchronized (managers) {
-            final var manager = managers.stream()
```
AsamK, line 100: Instead of always checking by number first and falling back to aci, you can check if the identifier is a uuid (`UuidUtil.INSTANCE.isUuid`) first. And then directly load by number or aci.

AsamK (approved):

Issue #1592, opened 2024-09-24
[Feature Request] Add stories compatibility


LunarN0v4:
Adding support for stories, including managing story groups would be useful.  
I'm making a bot which manages and creates stories for me, so I'd love to see this be added as a feature.

scottslewis:
Also would like to see this enhancement.  Would be willing to contribute to implementation and/or testing if desired.

AsamK:
The first step would be to design a cli interface for sending stories (maybe `sendStoryMessage`) and managing distribution lists for story groups (maybe `listStoryDistributionLists` and `updateStoryDistributionList`) with parameters for all the different functionalities.

LunarN0v4:
Is it required by the Signal protocol to actually have a distribution list?
Having a way to send stories without one would be nice to see (maybe via the recipients array?) for huge constantly-changing lists, but that's probably better off as a side-quest for the minute.

On Thursday, October 24, 2024 05:05 AEST, Sebastian Scheibner ***@***.***> wrote:

 
 
The first step would be to design a cli interface for sending stories (maybe sendStoryMessage) and managing distribution lists for story groups (maybe listStoryDistributionLists and updateStoryDistributionList) with parameters for all the different functionalities.
—
Reply to this email directly, view it on GitHub, or unsubscribe.
You are receiving this because you authored the thread.Message ID: ***@***.***>
[ { ***@***.***": "http://schema.org", ***@***.***": "EmailMessage", "potentialAction": { ***@***.***": "ViewAction", "target": "https://github.com/AsamK/signal-cli/issues/1592#issuecomment-2433214900", "url": "https://github.com/AsamK/signal-cli/issues/1592#issuecomment-2433214900", "name": "View Issue" }, "description": "View this Issue on GitHub", "publisher": { ***@***.***": "Organization", "name": "GitHub", "url": "https://github.com" } } ]

AsamK:
I would have to dig deeper into the code to check whether it's actually required. It will definitely be more efficient to use distribution lists when sending multiple stories to the same set of people, because it can reuse the sender key. 
For sending stories to an existing group it might not matter, but that needs to be checked.

LunarN0v4:
> I would have to dig deeper into the code to check whether it's actually required. It will definitely be more efficient to use distribution lists when sending multiple stories to the same set of people, because it can reuse the sender key. For sending stories to an existing group it might not matter, but that needs to be checked.

[...]

PR #2082, opened 2026-07-09
Add sendStory command for posting file attachment stories

tonycpsu:
## Summary

- Add `sendStory` command to post a file attachment story (image/video) to "My Story" (all contacts)
- Supports `--attachment` (required) and `--no-replies` (optional) arguments
- Works in both CLI and JSON-RPC modes
- Validates attachment MIME type (image/video only) before uploading
- Resolves eligible recipients before uploading to avoid wasted bandwidth on empty contact lists
- Sends multi-device sync transcript after delivery

Addresses #1592. Takes a different approach from #1713 — sends to `DistributionId.MY_STORY` (individual contacts via sender keys) rather than group stories, which avoids the group send endorsements blocker that stalled the prior attempt.

### CLI usage
```
signal-cli -a +1234567890 sendStory -a /path/to/image.jpg
signal-cli -a +1234567890 sendStory -a /path/to/video.mp4 --no-replies
```

### JSON-RPC usage
```json
{"jsonrpc":"2.0","method":"sendStory","id":"1","params":{"attachment":"/path/to/image.jpg"}}
```

### Scope

This is Phase 1 — file attachment stories to "My Story" only. Future phases could add:
- Text stories (styled text cards)
- Group story targeting
- Custom distribution list management

## Test plan

- [x] `:lib:compileJava` passes
- [x] `:compileJava` and `:compileTestJava` pass (full project)
- [x] `test` suite passes
- [x] Manual test: post an image story and verify it appears on recipients' devices
- [x] Manual test: post with `--no-replies` and verify reply is disabled
- [x] Manual test: attempt non-image/video attachment and verify rejection

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/internal/ManagerImpl.java
+++ b/lib/src/main/java/org/asamk/signal/manager/internal/ManagerImpl.java
@@ -110,7 +110,6 @@ import org.whispersystems.signalservice.api.messages.SignalServiceDataMessage;
 import org.whispersystems.signalservice.api.messages.SignalServicePreview;
 import org.whispersystems.signalservice.api.messages.SignalServiceReceiptMessage;
 import org.whispersystems.signalservice.api.messages.SignalServiceStoryMessage;
-import org.whispersystems.signalservice.api.messages.SignalServiceStoryMessageRecipient;
 import org.whispersystems.signalservice.api.messages.SignalServiceTypingMessage;
 import org.whispersystems.signalservice.api.messages.calls.AnswerMessage;
 import org.whispersystems.signalservice.api.messages.calls.BusyMessage;
@@ -119,7 +118,6 @@ import org.whispersystems.signalservice.api.messages.calls.IceUpdateMessage;
 import org.whispersystems.signalservice.api.messages.calls.OfferMessage;
 import org.whispersystems.signalservice.api.messages.calls.SignalServiceCallMessage;
 import org.whispersystems.signalservice.api.messages.multidevice.DeviceInfo;
-import org.whispersystems.signalservice.api.push.DistributionId;
 import org.whispersystems.signalservice.api.push.ServiceIdType;
 import org.whispersystems.signalservice.api.push.exceptions.CdsiResourceExhaustedException;
 import org.whispersystems.signalservice.api.util.DeviceNameUtil;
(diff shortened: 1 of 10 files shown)


lib/src/main/java/org/asamk/signal/manager/internal/ManagerImpl.java
```diff
@@ -826,6 +830,62 @@ public SendMessageResults sendEditMessage(
...
+        if (mimeType.isEmpty() || (!mimeType.get().startsWith("image/") && !mimeType.get().startsWith("video/"))) {
+            throw new AttachmentInvalidException(attachment,
+                    new IOException("Stories only support image and video attachments"));
+        }
+
+        final var recipients = account.getRecipientStore()
+                .getRecipients(true, Optional.of(false), Set.of(), Optional.empty());
+        final var recipientIds = recipients.stream()
+                .filter(r -> !r.getRecipientId().equals(account.getSelfRecipientId()))
+                .filter(r -> r.getContact() == null || !r.getContact().hideStory())
```
AsamK: Hm, I think you want `r.getContact() != null && !r.getContact().hideStory()` here, right?
Though it probably doesn't make a difference, as contacts are already filtered in the `getRecipients` call above.

lib/src/main/java/org/asamk/signal/manager/internal/ManagerImpl.java
```diff
@@ -826,6 +830,62 @@ public SendMessageResults sendEditMessage(
...
+                .sendStoryMessage(storyMessage, timestamp, recipientIds, allowsReplies);
+
+        final var storyMessageRecipients = sendResults.stream()
+                .filter(org.whispersystems.signalservice.api.messages.SendMessageResult::isSuccess)
+                .map(r -> new SignalServiceStoryMessageRecipient(r.getAddress(),
+                        List.of(DistributionId.MY_STORY.asUuid().toString()),
+                        allowsReplies))
+                .collect(Collectors.toSet());
+        try {
+            dependencies.getMessageSender().sendStorySyncMessage(storyMessage, timestamp, false, storyMessageRecipients);
```
AsamK: The `sendGroupStory` function already send a sync message internally, so doing it here again in redundant.

AsamK (dismissed):
Thanks, looks good.
Just the manual sending of the sync message can be removed, as that is already handled internally in libsignal-service.

lib/src/main/java/org/asamk/signal/manager/internal/ManagerImpl.java
tonycpsu: Good catch — fixed in 1799b434. The `getRecipients(true, ...)` call does already filter to contacts, but the corrected logic (`!= null && !hideStory()`) better expresses the intent.

lib/src/main/java/org/asamk/signal/manager/internal/ManagerImpl.java
tonycpsu: You're right — removed in 1799b434. Thanks for the review!

AsamK (approved):

PR #2084, opened 2026-07-10
Expose invalid envelope content in receive output

k-bx:
## Problem

Decrypted messages can fail `EnvelopeContentValidator`, for example when formatting or mention ranges extend beyond the UTF-16 message body.

signal-cli previously logged the validation reason and silently dropped the message. Downstream consumers received no structured event and could not associate the rejection with its sender, device, or envelope timestamps.

## Changes

This change:

- routes invalid-content results through the existing receive exception handler
- keeps the receive command running successfully
- includes sender, device, and envelope timestamps in the receive event
- adds a stable diagnostic code
- reports UTF-16 body length and offending range coordinates
- excludes the message body from diagnostic details
- logs the validator’s synthetic stack trace at debug level

Example:

```json
{
  "exception": {
    "message": "[DataMessage] Body range with out-of-bounds start/length!",
    "type": "InvalidEnvelopeContentException",
    "details": {
      "code": "DATA_MESSAGE_BODY_RANGE_OUT_OF_BOUNDS",
      "bodyLength": 10,
      "invalidBodyRanges": [
        {
          "index": 0,
          "start": 8,
          "length": 4,
          "type": "STYLE_BOLD"
        }
      ]
    }
  },
  "envelope": {
    "sourceUuid": "...",
    "sourceDevice": 2,
    "timestamp": 1234567890
  }
}
```

The malformed content remains rejected; only its diagnostic metadata is emitted. Existing exception JSON remains unchanged because `details` is omitted when unavailable.

## Testing

- Added coverage for extracting out-of-bounds body-range diagnostics.
- Added coverage for structured and ordinary JSON errors.
- Ran `./gradlew check` successfully.

### DIFF
--- a/lib/src/main/java/org/asamk/signal/manager/helper/IncomingMessageHandler.java
+++ b/lib/src/main/java/org/asamk/signal/manager/helper/IncomingMessageHandler.java
@@ -122,6 +122,8 @@ public final class IncomingMessageHandler {
         }
 
         SignalServiceContent content = null;
+        Content decryptedContent = null;
+        InvalidEnvelopeContentException validationException = null;
         if (!envelope.isReceipt()) {
             account.getIdentityKeyStore().setRetryingDecryption(true);
             try {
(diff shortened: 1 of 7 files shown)


lib/src/main/java/org/asamk/signal/manager/helper/IncomingMessageHandler.java
```diff
@@ -252,9 +294,105 @@ private SignalServiceContent validate(
...
+            }
+        } else if (dataMessage == null) {
+            body = null;
+            bodyRanges = null;
+        } else {
+            body = dataMessage.body;
+            bodyRanges = dataMessage.bodyRanges;
+        }
+
+        final Integer bodyLength = bodyRanges == null ? null : body == null ? 0 : body.length();
```
AsamK, line 322: `body.length` is not sufficient. Messages with more than 2000 chars store the body as an extra attachment.
I'd rather not have the additional length checks here at all. Just notifying the user that the ranges are incorrect should be enough.

PR #2098, opened 2026-08-07
Support the GroupsV2 "end group" (terminate) feature

FailSpy:
Signal recently added the ability for a group admin to permanently *terminate* a group ("End group" in the app UI): once it's terminated the group is read-only for everyone -- nobody, admins included, can send messages or start calls -- and members keep the existing history. Other clients silently drop any message sent to a terminated group. The pinned `signal-network` already carries the protobuf for it (`DecryptedGroup.terminated`, plus the `TerminateGroupAction` change), so signal-cli just wasn't reading or offering it yet. This wires it up both directions.

## Seeing that a group is terminated

Today signal-cli can't tell a terminated group from a live one, and it'll report a sent message as `SUCCESS` even though every other client drops it. This reads `DecryptedGroup.terminated` through `GroupInfo.isTerminated()` and:

- surfaces it in `listGroups` (JSON, and both the default and `--detailed` plaintext lines) and as the DBus `IsTerminated` property
- refuses to send to a terminated group locally -- messages, reactions, typing and group stories -- rather than send one other clients ignore
- drops incoming messages addressed to a terminated group (in `isNotAllowedToSendToGroup`, alongside the announcement-group handling)

The send refusal mirrors the existing `isAnnouncementGroup` wiring, with one difference: it's unconditional. Announcement-only exempts admins; a terminated group blocks everyone.

**On the `listGroups` output:** I put `Terminated` on the default (non-detailed) line, right after `Blocked`. `Blocked` is almost always false yet already lives there, so that line is a status-at-a-glance row, not an only-show-non-defaults one. Terminated is a peer of `Blocked` -- both mean "this isn't a usable conversation" -- and since a terminated group otherwise looks completely normal, the glanceable line is where it's most worth surfacing. Happy to move it behind `--detailed` if you'd rather leave the default line untouched.

## Terminating a group

Adds a `terminateGroup` command (admin-only, v2 groups) that sends the `TerminateGroupAction` via the library's own `createTerminateGroup()` and notifies the members. It follows the `quitGroup` shape end to end -- `GroupHelper.terminateGroup` -> `Manager`/`ManagerImpl` -> the `terminateGroup` command (command line + JSON-RPC) and the DBus `Group.terminateGroup` method -- with the same conflict-retry as `updateGroup`.

## Tests / scope

Added a unit test for the terminated model reads and the `Group` record wiring, in the existing JUnit style (no new deps). The send-path guards aren't unit-covered -- there's no existing test path for the outgoing/send layer to build on -- but the `isTerminated()` signal they rely on is.


lib/src/main/java/org/asamk/signal/manager/helper/GroupHelper.java
```diff
@@ -363,6 +363,28 @@ public SendGroupMessageResults updateGroup(
         return results;
     }
 
+    public SendGroupMessageResults endGroup(final GroupId groupId) throws IOException, GroupNotFoundException, NotAGroupMemberException {
```
AsamK: Call it `terminateGroup` to use consistent naming.
Same goes for all other new functions called `endGroup`

lib/src/main/java/org/asamk/signal/manager/helper/SendHelper.java
```diff
@@ -567,14 +567,18 @@ private List<SendMessageResult> sendGroupTypingMessage(
...
+    private GroupInfo getGroupForSending(GroupId groupId) throws GroupNotFoundException, NotAGroupMemberException, GroupSendingNotAllowedException {
         var g = context.getGroupHelper().getGroup(groupId);
         if (g == null) {
             throw new GroupNotFoundException(groupId);
         }
         if (!g.isMember(account.getSelfRecipientId())) {
             throw new NotAGroupMemberException(groupId, g.getTitle());
         }
+        if (g.isTerminated()) {
+            // The server accepts sends to an ended group but never delivers them, so refuse locally.
```
AsamK: Correction: The server does deliver messages to terminated groups, the server has no knowledge of which group the message belongs to, that's part of the encrypted content. Clients drop messages from terminated groups and don't show them.
Same for the comment in ManagerImpl.

AsamK (changes requested):
Thanks, looks good.
Just the naming is inconsistent, please use `terminate` instead of `end` everywhere

lib/src/main/java/org/asamk/signal/manager/helper/SendHelper.java
AsamK: Also signal-cli needs to drop incoming messages to terminated groups. Should be added in the `isNotAllowedToSendToGroup` method.

AsamK (approved):

PR #2117, opened 2026-09-13
Fix voice-call device ID for linked clients

d-b-w-gain:
## Summary

- pass the account's actual device ID to `signal-call-tunnel`
- retain device ID 1 for primary devices while correctly identifying linked devices
- add a focused regression test for tunnel configuration

## Motivation

`CallManager.buildConfig` currently hardcodes `local_device_id` to 1. When
signal-cli is linked to a primary Signal phone, its actual device ID is greater
than 1. RingRTC therefore receives the primary phone's identity for the
signal-cli endpoint and can fail media negotiation after the remote endpoint
answers.

This was observed while exercising the experimental outgoing voice-call path:
the call reached `RINGING_OUTGOING` and `CONNECTING`, then the tunnel ended with
`InternalFailure` after answer.

This PR is intentionally limited to the device-ID correctness issue. Handling a
call addressed to the same Signal account may require separate discussion.

## Validation

```text
./gradlew :lib:test --tests org.asamk.signal.manager.helper.CallManagerTest
BUILD SUCCESSFUL
```


AsamK:
@d-b-w-gain why is this PR a draft?

PR #2120, opened 2026-09-16
Add --attachment-dimensions and --attachment-blurhash to send

elShiaLabeouf:
## Summary

- add `send --attachment-dimensions` and `--attachment-blurhash`, matched to `--attachment` by position (`''` skips one)
- pass them through `Message` → `AttachmentHelper` → `AttachmentUtils`; image dimensions are still detected when none are given
- reject malformed values with a `UserErrorException` before uploading

## Why

Signal Desktop only shows a preview for a video that has [dimensions](https://github.com/signalapp/Signal-Desktop/blob/main/ts/util/Attachment.std.ts), and fills it with the attachment's BlurHash until the video is downloaded. signal-cli can't derive either from a video, so videos arrive as a plain file with a film icon.

A real thumbnail isn't an option: Desktop [ignores `AttachmentPointer.thumbnail`](https://github.com/signalapp/Signal-Desktop/blob/main/ts/textsecure/processDataMessage.preload.ts) on received attachments and makes its own from the file after download.

## Usage

```
signal-cli send +1234567890 -a clip.mp4 --attachment-dimensions 1080x1920 --attachment-blurhash 'LEHV6nWB2yk8pyo0adR*.7kCMdnj'
```

```json
{"jsonrpc":"2.0","method":"send","params":{"recipient":["+1234567890"],"attachments":["clip.mp4"],"attachmentDimensions":["1080x1920"],"attachmentBlurhash":["LEHV6nWB2yk8pyo0adR*.7kCMdnj"]},"id":1}
```

## Test plan

- [x] `./gradlew :test :lib:test` passes
- [x] `AttachmentUtilsTest`: supplied dimensions and BlurHash are set on the stream, and the bytes are unchanged
- [x] `SendCommandParsingTest`: CLI and JSON-RPC keep positions, and malformed values are rejected
- [x] Manual test: send a video with both options and check the placeholder in Signal Desktop before download:

#### Without this patch

Stock signal-cli: Signal Desktop shows its generic placeholder in a default-size box.
Screenshot From 2026-09-16 12-54-04

#### With this patch

With `--attachment-dimensions 1280x720` and a BlurHash: a blurred preview at the video's 16:9 shape, before download.
Screenshot From 2026-09-16 14-58-24

After clicking Download, Signal Desktop replaces the BlurHash with a thumbnail generated from the video.
Screenshot From 2026-09-16 14-59-38


lib/src/main/java/org/asamk/signal/manager/util/AttachmentUtils.java
AsamK: you can check `dimensions != null` here and skip probing completely in that case.

AsamK (commented):
Thanks, looks good. Just one comment to skip probing completely if dimensions are given by the user.

lib/src/main/java/org/asamk/signal/manager/util/AttachmentUtils.java
elShiaLabeouf: Done. Re-tested, works well:

image

Thanks for the quick review!

AsamK (approved):

PR #2123, opened 2026-09-17
Support linking phone-numberless Signal accounts

### LINKED ISSUE #2122: Support linking phone-numberless Signal accounts
When linking signal-cli to an account created without a phone number in Signal Android 8.28 beta, the Android app rejects the linking QR code with a message saying that the device is out of date and should be upgraded.

This uses the numberless account flow discussed in the [Android 8.28 beta thread](https://community.signalusers.org/t/beta-feedback-for-the-upcoming-android-8-28-release/76457).

### Reproduction

1. Create an account without a phone number in Signal Android 8.28 beta.
2. Generate a signal-cli device-link QR code. The original failure was observed through `bbernhard/signal-cli-rest-api:0.100-rootless`; the corresponding CLI entry point is `signal-cli link -n signal-cli`.
3. Scan the QR code from Android's Linked devices screen.

The phone rejects the link as an outdated device. Expected: signal-cli can link as a secondary device and subsequently identify the account by its ACI.

### Source investigation

At upstream commit `c146224493ad41a6f6ce14e17d75c590564b5c8f` (`0.14.9-SNAPSHOT`), the linking URI does not advertise `capabilities=nopni`. Provisioning also assumes that PNI and PNI key material are present, and account discovery assumes a phone number. Supporting this flow therefore needs both capability advertisement and handling of accounts without a phone number/PNI.

The requested scope is secondary-device linking and subsequent use through the CLI, JSON-RPC, and D-Bus. Primary registration or recovery using an Account Key is outside this request.

### Environment and evidence

- Mobile client: Signal Android 8.28 beta with a numberless account.
- Original gateway: `signal-cli-rest-api:0.100-rootless`.
- Source/build investigation: signal-cli `0.14.9-SNAPSHOT` at the upstream commit above; resolved `libsignal-client` dependency `0.102.1`.
- Local build verification: JVM distribution on Ubuntu 24.04 x86-64, using Temurin JDK `25.0.4.1+1` and `26.0.2.1+1`.

The original failure is the phone-side rejection described above; no scrubbed CLI failure log is attached. A patched build has successfully linked to the numberless account in a live test. I will follow this issue with the implementation and its automated regression tests.

emsi:
Closes #2122.

Accounts created without a phone number in Signal Android 8.28 beta cannot link to signal-cli: the link URI lacks `nopni`, and provisioning and account selection assume a phone number/PNI. This adds secondary-device support across the CLI, JSON-RPC, and D-Bus, using the ACI when no phone number exists.

### Changes

- Advertise `capabilities=nopni` and use the existing library's V2 linked-device registration API with ACI authentication and optional PNI keys. Validate the group-auth credential salt for numberless provisioning.
- Handle absent phone numbers/PNIs in account attributes, startup, group membership, sending, and receiving. Numberless accounts can be selected by ACI; recipient phone numbers must include an international country code.
- Add ACI information to JSON account output. For D-Bus, preserve numbered-account object paths, export numberless accounts by ACI, return an empty string for absent phone numbers, and add recipient enumeration that includes accounts without phone numbers.
- Mark numberless account-index entries explicitly so superseded numbered entries are not rediscovered. Resolve older ambiguous entries from saved account state, preserve ACI lookup for relinking, and retry classification when state is unavailable.
- Add regression tests, CLI/JSON-RPC/D-Bus documentation, and native-image metadata for the changed account record.

Primary numberless registration/recovery using an Account Key is outside this change. REST wrappers need their own handling for nullable phone numbers and ACI selectors.

### Validation

On Temurin JDK 25.0.4.1+1 and 26.0.2.1+1, Ubuntu 24.04 x86-64:

```sh
dbus-run-session -- ./gradlew --no-daemon cleanTest :libsignal-cli:cleanTest build jsonSchemas
```

Both runs passed: **193 tests**, with no failures or skips (51 application tests and 142 library tests). Coverage includes numbered and numberless provisioning, persisted accounts, superseded-account discovery, legacy index migration, locked/missing state, and D-Bus transport with mixed accounts and direct/group/self operations. D-Bus transport tests use an isolated session bus and a local backend; without a session bus, those transport tests are skipped.

Live linking and basic operation of the patched CLI were confirmed with the numberless Android account. Full live Signal messaging through D-Bus has not been verified.

All introduced Java changes were checked with upstream's IntelliJ formatter settings, and `git diff --check` passes. An additional check confirmed that upstream's older account-index record reads the new optional field without losing identifiers. Full reproducible container packaging, GraalVM native compilation, and upstream CI have not been run locally.


doc/numberless-accounts.md
```diff
@@ -0,0 +1,73 @@
+# Linking an account without a phone number
```
AsamK, line 1: Please don't add another documentation folder.
Instead a short description in the man page is enough. Further documentation can go into the github wiki.

lib/src/main/java/org/asamk/signal/manager/helper/AccountHelper.java
```diff
@@ -564,7 +570,22 @@ public void refreshDeviceName() throws IOException {
...
 
     public void updateAccountAttributes() throws IOException {
-        handleResponseException(dependencies.getAccountApi().setAccountAttributes(account.getAccountAttributes(null)));
+        if (account.getNumber() != null) {
+            handleResponseException(dependencies.getAccountApi()
+                    .setAccountAttributes(account.getAccountAttributes(null)));
+            return;
+        }
+        final var api = new AccountApiV2(dependencies.getAuthenticatedSignalWebSocket());
+        final RequestResult<Unit, ? extends SetAccountAttributesError> result = runSuspendBlocking(cont -> api.setAccountAttributes(
```
AsamK, line 579: You can use `handleResponseExceptionSuspend` here instead. I've extended it to throw a `BadRequestException` with the actual error.

lib/src/main/java/org/asamk/signal/manager/helper/AccountHelper.java
```diff
@@ -564,7 +570,22 @@ public void refreshDeviceName() throws IOException {
     }
 
     public void updateAccountAttributes() throws IOException {
-        handleResponseException(dependencies.getAccountApi().setAccountAttributes(account.getAccountAttributes(null)));
+        if (account.getNumber() != null) {
+            handleResponseException(dependencies.getAccountApi()
+                    .setAccountAttributes(account.getAccountAttributes(null)));
+            return;
+        }
+        final var api = new AccountApiV2(dependencies.getAuthenticatedSignalWebSocket());
```
AsamK, line 578: All API should be created as singletons in `SignalDependencies`, can be analogous to `getAccountApi`.

lib/src/main/java/org/asamk/signal/manager/internal/ProvisioningManagerImpl.java
```diff
@@ -296,6 +295,76 @@ public String finishDeviceLink(String deviceName) throws IOException, TimeoutExc
...
+            final PreKeyCollection aciPreKeys,
+            final PreKeyCollection pniPreKeys
+    ) throws IOException {
+        final var attrs = account.getAccountAttributesV2();
+        final var deviceAttributes = new RegistrationApiV2.DeviceAttributes(attrs.getFetchesMessages(),
+                attrs.getRegistrationId(),
+                attrs.getPniRegistrationId(),
+                attrs.getName(),
+                attrs.getCapabilities());
+        final RequestResult<LinkDeviceResponse, ? extends RegisterAsLinkedDeviceError> result = runSuspendBlocking(cont -> registrationApi.registerAsSecondaryDevice(
```
AsamK, line 327: Also use `handleResponseExceptionSuspend` here and use `try`/`catch` for error handling.

README.md
```diff
@@ -60,14 +61,16 @@ For a complete usage overview please read
 the [man page](https://github.com/AsamK/signal-cli/blob/master/man/signal-cli.1.adoc) and
 the [wiki](https://github.com/AsamK/signal-cli/wiki).
 
-Important: The ACCOUNT is your phone number in international format and must include the country calling code. Hence it
+For a numbered account, ACCOUNT is your phone number in international format and must include the country calling code. Hence it
 should start with a "+" sign. (See [Wikipedia](https://en.wikipedia.org/wiki/List_of_country_calling_codes) for a list
 of all country codes.)
+For a linked account without a phone number, use its ACI (Account ID) instead.
+See [numberless accounts](doc/numberless-accounts.md) for details.
```
AsamK, line 68: Should link to the wiki instead.

src/main/java/org/asamk/signal/dbus/DbusMultiAccountManagerImpl.java
```diff
@@ -45,10 +45,7 @@ public DbusMultiAccountManagerImpl(
 
     @Override
     public List<String> getAccountNumbers() {
```
AsamK, line 47: I think `getAccountNumbers` is now unused and could be removed.

src/main/java/org/asamk/signal/dbus/DbusManagerImpl.java
```diff
@@ -1073,6 +1082,23 @@ private String emptyIfNull(final String string) {
...
+            return new RecipientAddress(UUID.fromString(identifier));
+        } catch (IllegalArgumentException e) {
+            return new RecipientAddress(identifier);
+        }
+    }
+
+    private List<String> listRecipientIdentifiers() {
+        try {
+            return signal.listRecipientIdentifiers();
+        } catch (UnknownMethod e) {
```
AsamK, line 1096: This UnknownMethod check can be removed, signal-cli dbus client only supports the same version of the signal-cli dbus daemon.

lib/src/main/java/org/asamk/signal/manager/storage/accounts/AccountsStore.java
```diff
@@ -63,11 +63,14 @@ public synchronized Set<String> getAllNumbers() throws IOException {
     public synchronized Set<AccountsStorage.Account> getAllAccounts() throws IOException {
         return readAccounts().stream()
                 .filter(a -> a.environment() == null || serviceEnvironment.equals(a.environment()))
-                .filter(a -> a.number() != null)
+                .filter(a -> a.number() != null || (a.uuid() != null && Boolean.TRUE.equals(a.numberless())))
```
AsamK, line 66: I don't like adding the `numberless` field to the accounts file. Is it really necessary?
The check here could be just `.filter(a -> a.number() != null || a.uuid() != null`

AsamK (changes requested):
Thanks! Looks good in general, I've added a couple of comments.

lib/src/main/java/org/asamk/signal/manager/storage/accounts/AccountsStore.java
emsi: The initial implementation actually used the proposed `number != null || uuid != null` predicate. During regression testing, I've found that it incorrectly rediscovered superseded numbered accounts.
The current implementation was added to address that regression.
                                            
When `addAccount` or `updateAccount` assigns an existing phone number to another account, it clears the old index entry’s number but retains its UUID. That leaves `number = null and uuid != null`, the same combination as a genuine numberless account. The proposed predicate therefore includes both.    

The numberless flag distinguishes these cases while preserving the old ACI-to-path mapping used during relinking. For older ambiguous entries without the flag, the implementation checks the saved account state and persists the classification. If that state is locked or unavailable, classification is deferred and retried later.
               
I verified this again in an isolated copy: replacing only the filter with the proposed predicate makes both cases of `replacingANumberDoesNotRediscoverTheOldAccountAsNumberless` fail, covering both adding and updating an account. Both pass with the current implementation.      
               
I’m open to a simpler representation that avoids the additional field. However, the one-line replacement alone loses this distinction and reintroduces the regression.

AsamK: Ok, I'll merge as is and check if I find a simpler approach afterwards.

AsamK (approved):

PR #2126, opened 2026-09-20
man: update link section, signal-cli prints the QR code itself

### LINKED ISSUE #2125: man page: link section still says to generate the QR code with qrencode
The man page for `link` says to create a QR code from the URI with something like qrencode, but since 0.14.0 signal-cli already prints the QR code in the terminal when stdout is a tty. Worth updating the text so people don't go install qrencode for nothing.

nimble-turtle:
Fixes #2125. The link section told people to build a QR code with qrencode, but signal-cli renders one itself when attached to a terminal, so this flips the order and keeps qrencode as the fallback for non-tty use.


AsamK (approved):
Thanks!

